PR #80, opened 2025-09-22
Update `.gitignore`

flippingbits:
This PR adds more files and directories that should be excluded from Git to the `.gitignore` file


.gitignore
```diff
@@ -1,2 +1,42 @@
...
+
+# Other build byproducts
+*.pid
+local_mode_test.log
+coverage.txt
+.idea
+*.iml
+
+# servicebot workdir
+.servicebot/
```
bcscc, line 41:
```suggestion
.servicebot/

# IntelliJ Platform (plugin development)
.intellijPlatform/

# Gradle
.gradle/
build/
```
Should we also ignore the IntelliJ platform directory and gradle auto generated files?

> [The IntelliJ Platform cache directory should be excluded from the version control system. Add the .intellijPlatform entry to the .gitignore file.](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-tasks.html#verifyPlugin-verificationReportsFormats)

> [the .gradle directory should be marked as excluded from project indexing. And IDE does it automatically when creating a or importing new gradle project.](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360010661760-Should-gradle-be-excluded)

These are also included in the [IntelliJ Plugin template repo](https://github.com/JetBrains/intellij-platform-plugin-template/blob/main/.gitignore)'s gitignore file.

.gitignore
```diff
@@ -1,2 +1,42 @@
...
+.tmp
+.DS_Store
+target
+
+# Other build byproducts
+*.pid
+local_mode_test.log
+coverage.txt
+.idea
+*.iml
```
bcscc:
```suggestion
*.iml

# IntelliJ IDEA (User-specific and Gradlle generated on import files)
.idea/workspace.xml
.idea/shelf
.idea/dictionaries
.idea/modules.xml
.idea/gradle.xml
.idea/libraries/
```
NIT: Found these [JetBrains Support suggestion](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems) on the default .idea format for development sharing

flippingbits: Mh, is there any specific file in the `.idea` directory that you think is worth sharing via Git?

bcscc: I think we can ignore the directory entirely for now since there's no configurations that needs to be shared. We can always update it if we start setting some useful configs.

bcscc (approved):

PR #82, opened 2025-09-22
Manage Java and Gradle via SDKMAN!

flippingbits:
This change adds a `.sdkmanrc` file, so that we can manage Java and Gradle via [SDKMAN!](https://sdkman.io/) and streamline the installation of the prerequisites. Initially, the PR also removed the Gradle wrapper. Unfortunately, it caused issues for some users, so I decided to keep it for one.


eguicf (approved):

.sdkmanrc
```diff
@@ -0,0 +1,7 @@
+# Use SDKMAN! to run "sdk env install" to initialize with correct JDK version.
+#
+# Enable auto-env through setting sdkman_auto_env=true in ~/.sdkman/etc/config
+# See https://sdkman.io/usage#config
+
+gradle=8.14.3
+java=24.0.2-tem
```
bcscc, line 7: Should we also plan to follow up later to update dependences with the java version upgrade? Noticing some depreciation warnings when running development on 24.

flippingbits: Sounds good to me! Which warnings did you see?

PR #86, opened 2025-09-24
DTX-909: Setup make commands for initial test check pipeline in CI

bcscc:
## Summary of Changes

- Update `Makefile` with make targets for `./gradlew build/test/release-ci`
- Upgraded the project to use Gradle 8.4.1 across all relevant configuration files, including `build.gradle.kts` and `gradle-wrapper.properties`. Keeps the project version consistent with the version defined in `./sdkmanrc`.
- Configured the Semaphore CI workflow to install only Java via SDKMAN, Gradle installation is now delegated to the wrapper (`./gradlew`). This removes an redundant installation and shrinks the SDKMAN cache size from ~500MB to ~250MB. 
- Added caching for both Gradle and SDKMAN in new `mk-files/ci-build-cache.mk`, adapting logic from [ide-sidecar](https://github.com/confluentinc/ide-sidecar/blob/7640c2e752da8c28ae9e10f356e94b7331ded0e3/mk-files/semaphore.mk#L21-L54) to prevent stale cache uploads with a 7 day TIL.
   - Decided to cache only the `~/.gradle/caches` and `~/.gradle/wrapper` directories instead of the entire `~/.gradle` directory to avoid daemon and lock issues.
        * `~/.gradle/caches`: This includes the Dependency Cache and the Gradle Build Cache (for task outputs).
        * `~/.gradle/wrapper`: This caches the Gradle distribution itself, preventing repeated downloads. I found this approach from [semaphore-demo-android](https://github.com/semaphoreci-demos/semaphore-demo-android/blob/8b0b1653c7ea8d4c290bafd3a0ec8a0bab09b91d/.semaphore/semaphore.yml#L41-L43) and explained in the docs on [Gradle build anatomy](https://docs.gradle.org/current/userguide/gradle_directories_intermediate.html#gradle_user_home).
   - Cache keys are computed from a checksum of changes in `gradle.properties`, `build.gradle.kts`, and `gradle-wrapper.properties`. So cache is updated only when there are changes to these configs. The [Gradle build cache](https://docs.gradle.org/current/userguide/configuration_cache_enabling.html) (`org.gradle.caching=true`) is also enabled to reuse task outputs. 
- Updated the Semaphore pipeline to include the new cache `restore`/`store` operations. Unrelated prologue/epilogue steps were also removed.
- Enabled test reports in the CI pipeline. The existing JUnit tests run via `./gradlew test` and produce reports in `build/test-results/test/*TEST*.xml`. The workflow now merges each XML into a single report in CI.

## Any additional details or context that should be provided?
- Once merged, followup with a PR to confirm if caching is setup correctly and reduces build times for PR CI workflow check.


.semaphore/semaphore.yml
```diff
@@ -25,13 +25,16 @@ global_job_config:
   prologue:
     commands:
       - checkout
-      - make show-args
-      - . vault-setup
-      - make init-ci
+      # - make show-args
+      # - . vault-setup
+      - make ci-sem-cache-restore
+      - make setup-sdk
```
flippingbits, line 31: These changes will be overwritten by ServiceBot, which manages this section (see the comment in lines 1 to 6 in this file). [Similar to the ide-sidecar repo](https://github.com/confluentinc/ide-sidecar/blob/main/service.yml#L14), we could set the config `pipeline_enable` to `false` to prevent this.

Makefile
```diff
@@ -110,3 +110,27 @@ include ./mk-include/cc-vault.mk
...
+GRADLE = $(SDKMAN_INIT) && ./gradlew
+
+.PHONY: setup-sdk
+setup-sdk:
+	# Install SDKMAN! (https://sdkman.io/install/#ci-mode) if restored in cache already
+	@if [ ! -d "$(HOME)/.sdkman" ]; then \
+		curl -s "https://get.sdkman.io?ci=true&rcupdate=false" | bash; \
+	fi
+	# Install dependencies as defined in .sdkmanrc
+	$(SDKMAN_INIT) && sdk install java $$(grep "java=" .sdkmanrc | cut -d'=' -f2);
```
flippingbits, line 128: Should we run `sdk env install` instead, so that it installs Gradle, too?

```suggestion
	$(SDKMAN_INIT) && sdk env install
```

flippingbits (commented):
Thanks, @bcscc! I have two minor comments. Otherwise, your PR looks good to me.

Makefile
bcscc: Decided to install only Java via SDKMAN and have the Gradle installation delegated to the wrapper (`./gradlew`). This removes an redundant installation and shrinks the SDKMAN cache size from ~500MB to ~250MB. However, I mentioned in the pr:

> If we prefer SDKMAN for Gradle, I can switch back but wasn't sure if we want to keep the original repo's wrapper approach. [Gradle Wrapper Docs](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper)

Also quick followup on another question I had:
> Since the SDKMAN and Gradle versions do not change frequently, should we change their cache's TTL policy to be more than the current 7 day timeframe adapted from `ide-sidecar`?

flippingbits: > If we prefer SDKMAN for Gradle, I can switch back but thought it made sense to stick with the original repo's wrapper approach. [Gradle Wrapper Docs](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper)

Makes sense to me. Should we then remove Gradle from the `.sdkmanrc` file?

> Since the SDKMAN and Gradle versions do not change frequently, should we change their cache's TTL policy to be more than the current 7 day timeframe adapted from ide-sidecar?

Sounds good to me!

bcscc: > Makes sense to me. Should we then remove Gradle from the .sdkmanrc file?

I think it doesn't hurt to keep it for easier env/version management for development purposes. The CI workflow is already ignoring that line.

> Sounds good to me!

Ok, I will change it so it has no TTL policy and let Semaphore's [default cache manage policy](https://docs.semaphore.io/using-semaphore/cache): `For jobs running on Semaphore Cloud, the total cache size is 9.6GB. Older files are automatically deleted after 30 days or when the cache fills up.` However, feel free to let me know if you have another suggestion!

flippingbits (approved):
LGTM. Thank you, @bcscc!

Makefile
bcscc: Follow ticket for updating TTL: https://github.com/confluentinc/intellij/issues/91

PR #99, opened 2025-10-06
remove `com.intellij.modules.ultimate` dependency

shouples:
Removing the `ultimate` dependency should allow us to more quickly iterate without ensuring our licensed accounts are associated in the dev/sandbox window.

We shouldn't need to change this for now:
https://github.com/confluentinc/intellij/blob/af8a3637d0d6be631811d04307ae0519b16d3121/build.gradle.kts#L36

With `./gradlew clean runIde`, the plugin is available for use without going through the "Unlock Ultimate" steps:
image


flippingbits (approved):
LGTM, thanks!

PR #108, opened 2025-10-14
prototype: add telemetry opt-in setting

shouples:
Still rough early-prototype stage

image


resources/messages/KafkaBundle.properties
```diff
@@ -1046,3 +1046,8 @@ kerberos.settings.change.krb5.conf.comment=Restart the IDE to apply this setting
 status.unknown=Unknown
 auth.required.title=User authentication is required
 auth.error.authorize.link=Authorize
+# Telemetry consent settings
+settings.telemetry.consent.group.title=Telemetry
+settings.telemetry.consent.label=Send telemetry data to help improve the plugin
+settings.telemetry.consent.description=Allow sending anonymous usage data to help us improve the Kafka plugin experience
```
flippingbits, line 1052: I think the description should mention that the usage data is shared with us, i.e., `Confluent, Inc.`. WDYT?

PR #122, opened 2025-10-23
remove explicit `spring` and `microservices.jvm` dependencies in plugin.xml

shouples:
https://github.com/confluentinc/intellij/pull/118 didn't help much, since it still required we activate Ultimate in order to see the plugin appear. ~~Setting these non-JSON plugin dependencies to optional with dummy [configs](https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html#sample) seems to work fine and I can still open the topic Producer view.~~

Removing them entirely also seems to do the trick.


resources/META-INF/spring.xml
```diff
@@ -0,0 +1,7 @@
+<idea-plugin>
+    <extensions defaultExtensionNs="com.intellij">
+        <!--        <annotator-->
+        <!--                language="kotlin"-->
+        <!--                implementationClass="com.intellij.spring"/>-->
```
shouples: This doesn't work, but I have no idea what we would actually need to put here (if anything)

resources/META-INF/microservices-jvm.xml
```diff
@@ -0,0 +1,7 @@
+<idea-plugin>
+    <extensions defaultExtensionNs="com.intellij">
+        <!--        <annotator-->
+        <!--                language="kotlin"-->
+        <!--                implementationClass="com.intellij.spring"/>-->
```
shouples: Same as https://github.com/confluentinc/intellij/pull/122/files#r2455279255

bcscc (approved):
Thanks!

PR #139, opened 2025-11-05
Add test coverage for telemetry

### LINKED ISSUE #130: Add basic tests for telemetry setting behavior with `logUsage()`
At a minimum, we just need to make sure:
- no events are sent if the `enableUsageData` telemetry setting is **disabled**
- an event is sent (with the expected `TelemetryEvent`?) when the `enableUsageData` telemetry setting is **enabled**

bcscc:
## Summary of Changes

- **New test files** covering telemetry classes:
  - `TelemetryServiceTest.kt` - Tests core Segment analytics integration
  - `TelemetryTest.kt` - Tests wrapper functions and auth detection helper
  - `ActionTelemetryListenerTest.kt` - Tests automatic action tracking
- Changed some methods/properties from `private` to `internal` in `TelemetryService`, `ActionTelemetryListener`, and `Telemetry.kt` to enable direct testing without reflection
- Changed `TelemetryEvent` from `sealed class` to `internal interface` to allow creating a test instance
- Added `mockito-kotlin = "5.4.0"` for Kotlin-friendly mocking

1. Run tests: `./gradlew test --tests "io.confluent.intellijplugin.telemetry.*"`
2. View test report: `open build/reports/tests/test/index.html`
3. Verify all tests pass

- Fixes #130 

## Pull request checklist

#### Tests

- [X] Added new
- [ ] Updated existing
- [ ] Deleted existing


test/io/confluent/intellijplugin/telemetry/TelemetryServiceTest.kt
```diff
@@ -0,0 +1,148 @@
+package io.confluent.intellijplugin.telemetry
+
+import com.intellij.openapi.application.ApplicationInfo
+import com.intellij.openapi.util.SystemInfo
```
eguicf: Thanks for the PR! Click-tested and it works good.
These 2 imports seemed not used in this file.

bcscc: Thanks for the catch!

src/io/confluent/intellijplugin/telemetry/Telemetry.kt
eguicf: I just realized and it might be related to previous Telemetry PR: this line would return "SASL" early if both SASL and SSL are present, where the correct return should be something like "SASL_SSL".

bcscc: Fixed, thanks!

gradle/libs.versions.toml
```diff
@@ -9,6 +9,7 @@ kotlin-metadata = "2.2.20"
 junit-jupiter = "5.10.0"
 junit-jupiter-params = "5.10.2"
 junit4 = "4.13.2"
+mockito-kotlin = "5.4.0"
```
flippingbits, line 12: Should we use the most recent release of `mockito-kotlin`?


```suggestion
mockito-kotlin = "6.1.0"
```

flippingbits (approved):
Great work, @bcscc! I have only one question about the version of `mockito-kotlin`.

src/io/confluent/intellijplugin/telemetry/Telemetry.kt
```diff
@@ -27,10 +27,14 @@ fun logUser(traits: Map<String, Any>) {
  */
 fun determineAuthMethod(connectionData: KafkaConnectionData): String {
     val properties = connectionData.secretProperties
+    val hasSasl = properties.contains("sasl.mechanism")
+    val hasSsl = properties.contains("ssl.keystore")
```
flippingbits, line 31: nit: It'd be great if we could also look at the client property [security.protocol](https://kafka.apache.org/41/javadoc/org/apache/kafka/common/security/auth/SecurityProtocol.html), which must be set to `SASL_PLAINTEXT` or `SASL_SSL` when using SASL for authentication and must be set to `SASL_SSL` or `SSL` when using SSL, but it's probably quite complicated because it's implemented as a string, so we'd have to parse it.

gradle/libs.versions.toml
bcscc: I choose 5.4.0 because I assume it's the most stable since it has the highest [usage](https://mvnrepository.com/artifact/org.mockito.kotlin/mockito-kotlin). However, I can also update it and test if you think it's ok to use the most recent!

src/io/confluent/intellijplugin/telemetry/Telemetry.kt
bcscc: Made an issue: https://github.com/confluentinc/intellij/issues, will look into it!

gradle/libs.versions.toml
flippingbits: Got it! Maybe its usage is the highest because it's around for a longer time. We can also update at a later point.

PR #178, opened 2025-11-14
chore: set up CI pipeline for releasing plugin versions

shouples:
## Summary of Changes

This PR does two main things:
- sets up some initial versioning data points:
  - current version in `plugin.xml` ([docs](https://plugins.jetbrains.com/docs/intellij/plugin-configuration-file.html#idea-plugin__version))
  - under-development version in `.next/version.txt`
- a `release-version` pipeline that follows the same structure as the VS Code extension:
  - VS Code extension release-version pipeline: https://github.com/confluentinc/vscode/blob/main/.semaphore/release-version.yml
  - and its servicebot definition for the (manual) Semaphore task: https://github.com/confluentinc/vscode/blob/dc4c69257631b2c9a1e0ac0d8cd27c1893878c40/service.yml#L18-L45

Sample task run (https://semaphore.ci.confluent.io/workflows/c5ad75e3-f453-481d-9815-becf744896cd?pipeline_id=49602e8c-5ba5-4ee9-abb3-6ad294c26631):
image

Using a placeholder task created in the Semaphore UI while leaving some of the parameters optional to ensure it could run off of `main` if we eventually turn this into a scheduled prerelease:
image
(This PR's branch was used for the run-branch since it included the pipeline file, as well as the target branch since it included in the version used for creating the release.)

> [!NOTE]
> The GitHub prerelease from the above setup can be found here: https://github.com/confluentinc/intellij/releases/tag/v253.25909.0-pre

- The JetBrains [docs](https://plugins.jetbrains.com/docs/intellij/plugin-configuration-file.html#idea-plugin__version) hint at using `patchPluginXml` instead of explicitly setting `<version>`, but I'm not sure if we need that or if it would work well with our expected release flow.
  > Required **yes**; ignored in an [additional config file](https://plugins.jetbrains.com/docs/intellij/plugin-configuration-file.html#additional-plugin-configuration-files)
The element can be skipped in the source `plugin.xml` file if the Gradle plugin `patchPluginXml` task ([2.x](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-tasks.html#patchPluginXml), [1.x](https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-patchpluginxml)) is enabled and configured.

## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


resources/META-INF/plugin.xml
```diff
@@ -1,6 +1,7 @@
 <idea-plugin>
     <id>com.intellij.bigdatatools.kafka</id>
     <name>Kafka</name>
+    <version>253.25908.13</version>
```
shouples: Based on version released in October: https://plugins.jetbrains.com/plugin/21704-kafka/versions, but maybe we want to specify something else here?

build.gradle.kts
```diff
@@ -191,6 +191,15 @@ tasks {
...
         useJUnitPlatform()
     }
+
+    patchPluginXml {
+        val releaseName = System.getenv("RELEASE_NAME")
+        // if RELEASE_NAME is set (e.g. from a release job in CI), patch it in plugin.xml
+        // so the resulting plugin zip has the correct version number when installed
+        if (!releaseName.isNullOrEmpty()) {
+            version = releaseName.removePrefix("v")
+        }
```
shouples, line 201: This is slightly different (and I'd argue more correct) compared to what the VS Code extension does: over there, we have an automated process update the "current" version in [package.json](https://github.com/confluentinc/vscode/blob/b573ee06b95f764b1ee0f5284490b7a6370e6649/package.json#L6) to be the [`.versions/next.txt`](https://github.com/confluentinc/vscode/blob/b573ee06b95f764b1ee0f5284490b7a6370e6649/.versions/next.txt#L1) value with a microversion suffix. I don't know if we want to follow the same path there, but I think any adjustments to handling automatic version updates are outside the scope of this PR.

.semaphore/release-version.yml
```diff
@@ -0,0 +1,147 @@
...
+                echo "Using provided RELEASE_NAME: $RELEASE_NAME"
+              else
+                VERSION=$(grep '<version>' resources/META-INF/plugin.xml | sed -E 's/.*<version>([^<]*)<\/version>.*/\1/' | xargs)
+                echo "Extracted version from plugin.xml: $VERSION"
+                RELEASE_NAME="v$VERSION"
+              fi
+              if [ "$IS_PRERELEASE" = "true" ]; then
+                if [[ "$RELEASE_NAME" != *"-pre"* ]]; then
+                  RELEASE_NAME="${RELEASE_NAME}-pre"
+                fi
```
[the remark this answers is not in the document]
shouples: This is the [same setup](https://github.com/confluentinc/vscode/blob/21a7bb2d82e9fc79e0cf71b96ebb8722a940088a/.semaphore/release-version.yml#L61-L65) we use in the `vscode` repo and we haven't had any issues with it, so I'm inclined to leave it for consistency.

.semaphore/release-version.yml
```diff
@@ -0,0 +1,147 @@
...
+      jobs:
+        - name: "Parameter Verification"
+          commands:
+            # Ensure RELEASE_NAME is set and adjust for prereleases, if necessary
+            - |
+              if [ -n "$RELEASE_NAME" ] && [ "$RELEASE_NAME" != "RELEASE_NAME" ]; then
+                echo "Using provided RELEASE_NAME: $RELEASE_NAME"
+              else
+                VERSION=$(grep '<version>' resources/META-INF/plugin.xml | sed -E 's/.*<version>([^<]*)<\/version>.*/\1/' | xargs)
+                echo "Extracted version from plugin.xml: $VERSION"
```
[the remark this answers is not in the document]
shouples: I don't think we can go off of `.versions/next.txt` unless we have a better handle on exactly what the `<version>` in `plugin.xml` should represent. @flippingbits what do you think?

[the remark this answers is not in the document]
flippingbits: Going off the version in `plugin.xml` makes sense to me. For pre releases, the version in `plugin.xml` could be based on the one from `.versions/next.txt` with a microversion suffix. WDYT?

.semaphore/release-version.yml
```diff
@@ -0,0 +1,147 @@
...
+            - git push origin --delete "$RELEASE_NAME" || true
+  - name: "Create GitHub Release and Tag"
+    dependencies: ["Delete GitHub Release and Tag"]
+    task:
+      jobs:
+        - name: "Create GitHub Release and Tag"
+          commands:
+            - artifact pull workflow RELEASE_NAME.txt
+            - RELEASE_NAME=$(cat RELEASE_NAME.txt)
+            # Try to determine previous release tag for generating release notes
```
flippingbits, line 77: I think we want to manage the [change notes](https://github.com/confluentinc/intellij/blob/main/resources/META-INF/plugin.xml#L193-L222) in `plugin.xml`, so that they show up on the Marketplace. I'm wondering if we should parse them here and add them to the GitHub release, too. What do you think about it?

resources/META-INF/plugin.xml
flippingbits: I think this should not point to the version of the most recent release, since it's likely that the codebase has changed since then, but the version from `.versions/next.txt` with a microversion suffix.

build.gradle.kts
flippingbits: I feel it'd make sense to use the same approach but we can treat it as a follow up if you like?

flippingbits (commented):
Great work, @shouples! I only have a question on what version we should store in `plugin.xml`. Otherwise, your PR looks good to me.

.semaphore/release-version.yml
shouples: That would be a nice enhancement! Maybe that can be a follow-up to this PR and we can set up a similar PR extension-side for parsing from its `CHANGELOG.md` file?

flippingbits: Sounds good to me! Taking care of it in a follow-up PR makes sense.

.semaphore/release-version.yml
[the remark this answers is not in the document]
shouples: Yeah, I think making the "current" version just the `next.txt` version with a microversion suffix makes sense.

resources/META-INF/plugin.xml
```diff
@@ -2,6 +2,7 @@
     <id>com.intellij.bigdatatools.kafka</id>
     <name>Kafka</name>
     <vendor>Confluent</vendor>
+    <version>253.25909.0</version>
```
flippingbits: nit: Should we add the suffix `-pre` to the version to make clear it's a pre-release?

flippingbits (dismissed):
Thanks for the update, @shouples! I have one more minor question, which should not block the PR.

flippingbits (approved):

Issue #153, opened 2025-11-07
AWS Region Error when creating a new connection


bcscc:
Error when creating a new connection. Isolated to when we upgraded [`glue-schema`](https://github.com/confluentinc/intellij/commit/206a700016b95689359c767ca804aedd18e27063#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfL6). However upgrade was needed because of unit testing failing introduced in [commit](https://github.com/confluentinc/intellij/commit/4e51cd30459dda28f93aba18504ac7f708ab9e35).

```
2025-11-07 12:51:13,426 [  32914] SEVERE - #io.confluent.intellijplugin.aws.settings.models.AwsRegionGroup$Companion - We have not classified AWS Regions [ca-west-1]
java.lang.Throwable: We have not classified AWS Regions [ca-west-1]
        at com.intellij.openapi.diagnostic.Logger.error(Logger.java:375)
        at io.confluent.intellijplugin.aws.settings.models.AwsRegionGroup.<clinit>(AwsRegionGroup.kt:54)
        at io.confluent.intellijplugin.aws.ui.AwsComponentsBuilder.getAwsRegions(AwsComponentsBuilder.kt:9)
        at io.confluent.intellijplugin.aws.ui.external.AwsSettingsComponentForKafka.getComponentRows$lambda$22$lambda$2(AwsSettingsComponentForKafka.kt:54)
        at com.intellij.ui.dsl.builder.impl.PanelImpl.row(PanelImpl.kt:59)
        at com.intellij.ui.dsl.builder.impl.PanelImpl.row(PanelImpl.kt:45)
        at com.intellij.ui.dsl.builder.impl.PanelImpl.row(PanelImpl.kt:18)
        at io.confluent.intellijplugin.aws.ui.external.AwsSettingsComponentForKafka.getComponentRows(AwsSettingsComponentForKafka.kt:53)
...
2025-11-07 12:51:13,428 [  32916] SEVERE - #io.confluent.intellijplugin.aws.settings.models.AwsRegionGroup$Companion - Plugin to blame: Kafka version: 252.23892.409
2025-11-07 12:51:13,428 [  32916] SEVERE - #c.i.s.ComponentManagerImpl - io.confluent.intellijplugin.aws.settings.models.AwsRegionGroup <clinit> requests com.intellij.ide.plugins.PluginUtil instance. Class initialization must not depend on services. Consider using instance of the service on-demand instead.
[...]

PR #211, opened 2025-11-20
Add ca-west-1 to list of supported AWS regions

flippingbits:
## Summary of Changes

This change adds `ca-west-1` to the list of supported AWS regions and makes sure we don't throw an error when we stumble upon an unknown AWS region.

Fixes #153

Create new connection **immediately** after launching IntelliJ and make sure you don't see any error being thrown.

-

## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


src/io/confluent/intellijplugin/aws/settings/models/AwsRegionGroup.kt
```diff
@@ -51,7 +53,7 @@ enum class AwsRegionGroup(override val id: String, override val title: String) :
             val otherRegions =
                 Region.regions() - globalRegions - usGovRegions - chinaRegions - globalRegions - specialGlobalRegion
             if (otherRegions.isNotEmpty()) {
-                thisLogger().error("We have not classified AWS Regions $otherRegions")
+                thisLogger().warn("We have not classified AWS Regions $otherRegions")
```
shouples, line 56: Confirmed there is no error pop-up with just this change applied to `main` ✅ 

Image

shouples (commented):
This works for me! I do see some other regions in the warning log though - should those be added in this branch as well?

src/io/confluent/intellijplugin/aws/settings/models/AwsRegionGroup.kt
bcscc: ^ same missing regions on my end

flippingbits: Interesting! They did not show up on my end. Did you use the IntelliJ 25.3 beta for testing the change?

shouples (approved):
Thanks @flippingbits!

PR #246, opened 2025-12-11
feat(OAuth): Create Http client methods and requests (2/5)

bcscc:
## Summary of Changes

- Implement `CCloudOAuthHttpClient` using [IntelliJ's HTTPs API](https://github.com/JetBrains/intellij-community/tree/1d6afa5823b6d2c7451ee75cfac38201f8c312dc/platform/ide-core/src/com/intellij/util/io) for OAuth HTTP operations
  - Chose IntelliJ's built-in HTTP to avoid additional dependencies and better IDE integration 
- Wire up all token exchange methods in `CCloudOAuthContext`:
  - `exchangeAuthorizationCode()` - POST form data to Auth0 `/oauth/token` endpoint
  - `createTokensFromRefreshToken()` - POST form data to refresh tokens
  - `exchangeControlPlaneToken()` - POST JSON to `/api/sessions`, extract `auth_token` from Set-Cookie header
  - `exchangeDataPlaneToken()` - POST JSON with Bearer auth to` /api/access_tokens`
  - `checkAuthenticationStatus()` - GET with Bearer auth to `/api/check_jwt`
- Fix `CCloudOAuthErrors.kt` where `@SerialName` annotations were mismatched with field names

### Feature Branch Stack (feature/CCloudOAuth)
- #232
  - #246 **(this)**
    - #247 
      - #251 
        - #252

### Optional: Click-testing instructions

1.

### Optional: Any additional details or context that should be provided?

- Added initial test coverage for `extractCookie` helper, will follow up with more compressive test coverage for all methods

## Pull request checklist

Please check if your PR fulfills the following (if applicable):

#### Tests

- [X] Added new
- [ ] Updated existing
- [ ] Deleted existing

PR #232, opened 2025-11-26
feat(OAuth): Create core OAuth structures (1/5)

bcscc:
## Summary of Changes

- Add core CCloud OAuth 2.0 authentication structures based on [ide-sidecar implementation](https://github.com/confluentinc/ide-sidecar/tree/main/src/main/java/io/confluent/idesidecar/restapi/auth)
- Add `kotlinx-serialization-json` dependency for response JSON [data serialization](https://kotlinlang.org/docs/serialization.html) (v1.9.0) ([docs](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/))
  - Chose over [Jackson usage in ide-sidecar](https://github.com/confluentinc/ide-sidecar/blob/0dd664ea14a5c89e0f968d9c0a38b0351d26829a/src/main/java/io/confluent/idesidecar/restapi/auth/CCloudOAuthContext.java#L783-L902) for native Kotlin null handling and better performance. Note: compared with [Moshi](https://medium.com/@kacper.wojciechowski/moshi-vs-kotlinx-serialization-the-ultimate-benchmark-a7ed776a46c0) and [Jackson](https://www.ericthecoder.com/2020/11/23/benchmarking-kotlin-json-parsers-jackson-kotlin-and-kotlinx-serialization/)
- Use Kotlin `suspend` + `Result<T>` instead of Vert.x `Future<T>` for async operations to better align with modern [IntelliJ Plugin SDK coroutine APIs](https://plugins.jetbrains.com/docs/intellij/kotlin-coroutines.html)
- `CCloudOAuthContext` modeled after ide-sidecar's [`CCloudOAuthContext.java`](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/auth/CCloudOAuthContext.java)
  - Implements same OAuth PKCE flow with code challenge/verifier and state parameter 
  - Preserved same token lifecycle (auth code → ID token → control plane → data plane) and thread-safety with `ReentrantReadWriteLock`
- `CCloudOAuthTokens` modeled after ide-sidecar's [`Tokens`](https://github.com/confluentinc/ide-sidecar/blob/0dd664ea14a5c89e0f968d9c0a38b0351d26829a/src/main/java/io/confluent/idesidecar/restapi/auth/CCloudOAuthContext.java#L636-L781) class for immutable state updates to tokens
[...]

PR #247, opened 2025-12-11
feat(OAuth): Create callback web server (3/5)

bcscc:
## Summary of Changes

- Implement `CCloudOAuthCallbackServer` using Java's built-in [`HttpServer`](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.httpserver/com/sun/net/httpserver/HttpServer.html) for OAuth callback handling
  - Chose Java's built-in HTTP server to avoid additional dependencies and maintain simplicity
  - Binds to localhost address (`127.0.0.1`) on configured callback port
- Handle OAuth callback flow:
  - Validate `state` parameter and extract authorization `code` from callback URL
  - Handle `error` parameter from Auth0 when authentication fails
  - Trigger token exchange via `CCloudOAuthContext.createTokensFromAuthorizationCode()`
- Create HTML templates for callback pages
- Callback server automatically stops after handling one callback (success or failure)

### Feature Branch Stack (feature/CCloudOAuth)
- #232
  - #246 
    - #247 **(this)**
      - #251 
        - #252

1.

- Fixes #151 
- Added test coverage for `parseQueryString` helper, will add more test coverage in follow-up PR.

## Pull request checklist

#### Tests

- [X] Added new
- [ ] Updated existing
- [ ] Deleted existing


resources/templates/styles.html
```diff
@@ -0,0 +1,49 @@
+<style>
```
flippingbits, line 1: Should this be a `.css` file?

src/io/confluent/intellijplugin/ccloud/auth/CCloudOAuthCallbackServer.kt
```diff
@@ -0,0 +1,196 @@
...
+import java.net.InetAddress
+import java.net.InetSocketAddress
+
+/**
+ * HTTP server to receive OAuth callback from Confluent Cloud.
+ *
+ * Flow:
+ * 1. User clicks "Sign In" → browser opens Auth0 authorize URL
+ * 2. User authenticates on Confluent Cloud
+ * 3. Auth0 redirects to: http://localhost:26636/gateway/v1/callback-intellij-docs?code=XXX&state=YYY
```
flippingbits: nit:


```suggestion
 * 3. Auth0 redirects to: http://localhost:26638/gateway/v1/callback-intellij-docs?code=XXX&state=YYY
```

src/io/confluent/intellijplugin/ccloud/auth/CCloudOAuthCallbackServer.kt
```diff
@@ -0,0 +1,196 @@
...
+                    .onSuccess { context ->
+                        sendResponse(exchange, 200, successHtml(context.getUserEmail()))
+                        CoroutineScope(Dispatchers.IO).launch {
+                            onSuccess(context)
+                            stop()
+                        }
+                    }
+                    .onFailure { exception ->
+                        val message = exception.message ?: "Token exchange failed"
+                        sendResponse(exchange, 400, errorHtml(message))
```
flippingbits: A failed token exchange smells a bit like a server-side failure, so maybe return the status code `500` instead?

src/io/confluent/intellijplugin/ccloud/auth/CCloudOAuthCallbackServer.kt
```diff
@@ -0,0 +1,196 @@
...
+                    if (parts.size == 2) parts[0] to java.net.URLDecoder.decode(parts[1], "UTF-8")
+                    else null
+                }
+                .toMap()
+        }
+    }
+
+    /**
+     * Start the callback server on the configured port.
+     * Server will automatically stop after handling one callback.
```
flippingbits, line 72: Would we start it again when the user needs to authenticate with Confluent Cloud another time?

resources/templates/styles.html
bcscc: It can be, but since it will have to be injected and wrapped with `<style>` anyway into the callback htmls, I thought it might be slightly simpler to keep it an inline html. Happy to change it, if you think a `.css` file will be better!

src/io/confluent/intellijplugin/ccloud/auth/CCloudOAuthCallbackServer.kt
bcscc: Yes, the plan was to keep it contained for each authentication attempt. I set it up this way to avoid leaving a long-running server, but I know sidecar does this. If you think there's a case I'm missing where we may need an active server, can change it to always run.

flippingbits: Fair point! The only use case that I can think of is exposing a special unauthenticated endpoint that could be used to check the existence of the plugin. For instance, if an external UI, like the Confluent Cloud Console, wants to check if the user has the plugin installed, they could call this endpoint. I think we can keep it as a follow up.

resources/templates/styles.html
flippingbits: I don't have a strong opinion about it. I think both options are fine. Feel free to keep it as is.

flippingbits (approved):

PR #320, opened 2026-01-30
minor: set a preferred size for the details fields to accommodate scrollbar

### LINKED ISSUE #227: Message Values are unreadable in "Details" panel
Long values introduce a scrollbar that obstructs the value underneath.

Image

noeldevelops:
## Summary of Changes

- Fixes https://github.com/confluentinc/intellij/issues/227 
- The key and value field details components now have a starting height of 50 pixels, which provides enough vertical space to accommodate the horizontal scrollbar when a long string appears
-  The width is set to 400 pixels as a default; both width and height will expand as needed via the layout manager                                                         

AFTER: 

https://github.com/user-attachments/assets/4442fa7a-023b-4a0d-92aa-c2b6a5b9d3df

1. Start a new topic consumer in the plugin, but set the schema to Base64 instead of the correct schema for the values (you can choose a topic with SR or JSON schema messages already there, or you can use the Producer feature to send a message to the topic with a different schema). You can stop the Consumer after the message comes in- doesn't have to be running. 
2. Double-click the row with the mismatched message to open the right Details panel. 
3. The value should be visible even if there is a horizontal scrollbar - no overlap of scrollbar & text. 

- Tried a rabbit hole to get the underlying component to make room for the scrollbar automatically but it was not clear how or why. Need more [Java Swing](https://docs.oracle.com/javase/tutorial/uiswing/layout/index.html) study.  
- This component is used in multiple panels so I kept the change scoped only to the Details; not sure if there are other places that need more height. 
- **The next level of support discussed as a team would be to add a warning if users are trying to consume with the wrong/mismatched schema, if possible.** Tracked [here](https://github.com/confluentinc/intellij/issues/321)

## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


flippingbits (approved):
LGTM. Thanks, @noeldevelops!

PR #330, opened 2026-02-03
[main] chore: update third party notices

flippingbits:
[main] chore: update third party notices


bcscc (approved):
LGTM, same result from running locally

PR #331, opened 2026-02-03
fix(CI): add `sem-version java 21` for third party notices

shouples:
`fossa-cli` errors were silently warning in the background (only visible with `--debug` added) and hinted at something in our Java setup (in CI) not being configured correctly:
```
FAILURE: Build failed with an exception.
          * What went wrong:
          A problem occurred configuring root project 'intellij'.
          > Could not resolve all artifacts for configuration 'classpath'.
             > Could not resolve org.jetbrains.intellij.platform:intellij-platform-gradle-plugin:2.9.0.
               Required by:
                   root project : > org.jetbrains.intellij.platform:org.jetbrains.intellij.platform.gradle.plugin:2.9.0
                > Dependency requires at least JVM runtime version 17. This build uses a Java 8 JVM.
          * Try:
          > Run this build using a Java 17 or newer JVM.
          > Run with --stacktrace option to get the stack trace.
          > Run with --info or --debug option to get more log output.
          > Run with --scan to get full insights.
          > Get more help at https://help.gradle.org/.
          BUILD FAILED in 22s
```
After a few iterations of debugging and attempted workarounds, I noticed `ide-sidecar` set `sem-version java 21`: https://github.com/confluentinc/ide-sidecar/blob/bded9700f639c7b7ce0d0dd3c835cc790a124113/.semaphore/third-party-notices.yml#L18

Once that was set here, the (temporary) task to generate the third party notices report worked and generated a full THIRD_PARTY_NOTICES.txt as we'd expect.

Previous failing run: https://semaphore.ci.confluent.io/jobs/e0f3ecc7-6a06-442c-b184-990416914c96#L230-L233

Successful run: https://semaphore.ci.confluent.io/jobs/3e110371-8b72-4cf0-8a04-0fe93a308ea2


scripts/generate-third-party-notices.sh
```diff
@@ -19,15 +19,21 @@ retry() {
 main() {
   # Verify that fossa CLI is available (should be natively installed in CI agent)
   command -v fossa || { echo "fossa CLI not found in PATH"; exit 1; }
-  
+
+  fossa --version
+  java -version
```
shouples, line 24: Could remove these, but they may be helpful for any future debugging if things go awry again

scripts/generate-third-party-notices.sh
```diff
@@ -19,15 +19,21 @@ retry() {
...
-  fossa analyze --exclude-path build --only-target gradle
-  
+
+  # Keep --debug to show detailed logs for troubleshooting in the event we see weird reports that
+  # don't exit with an error code
+  fossa analyze --debug --exclude-path build --only-target gradle
+
   # This might timeout so retry a few times
-  retry "fossa report attribution --format text > THIRD_PARTY_NOTICES.txt" 3
+  retry "fossa report attribution --debug --format text > THIRD_PARTY_NOTICES.txt" 3
```
flippingbits, line 36: Did you check that adding the `--debug` flag doesn't alter the output written to `THIRD_PARTY_NOTICES.txt`?

flippingbits (approved):
Thanks for the fix 🥇 LGTM.

scripts/generate-third-party-notices.sh
shouples: Yep, from local testing I've only seen it create a `fossa.telemetry.json` file and show debug-level logs while running

PR #346, opened 2026-02-05
add `intellij-sdk` claude skill

shouples:
Similar to the `vscode-api` skill (https://github.com/confluentinc/vscode/pull/3226), this adds a new `intellij-sdk` skill to dynamically fetch documentation for Intellij plugin development.

Sample use:
image


.claude/skills/intellij-sdk/SKILL.md
```diff
@@ -0,0 +1,232 @@
+---
+name: intellij-sdk
+description:
+  IntelliJ Platform SDK lookup for plugin development. Use when the user asks about IntelliJ Platform
+  APIs, extension points, threading rules, PSI, VFS, services, UI components, or how to implement
+  plugin features. Triggers on questions like "how do I create a tool window", "what's the API for X",
+  "IntelliJ threading rules", or checking platform API compatibility.
+user-invocable: false
```
shouples, line 8: This just means we wouldn't be able to run `/intellij-sdk` directly - do we want that kept enabled?

bcscc: so it can only be invoked by claude when instructed?

shouples: Yeah, claude would have to figure out when to use it based on the conversation _or_ we would have to say something to the effect of "use intellij-sdk to check ___"

bcscc: I see, maybe worth setting it to true? That way it can also be used as a form of code review?

bcscc (approved):

PR #355, opened 2026-02-09
MINOR: make run configs more explicit and add default project path

shouples:
Two main changes here:
- now the "Run Plugin" (`gradle runIde`) and "Run Plugin (Clean Build)" (`gradle clean runIde`) configs are separate
- `IDEA_PROJECT_PATH` to define the launch directory so we don't have to click through the Welcome/projects window each time we use `runIde`

https://github.com/user-attachments/assets/280baf91-f234-46ed-bd83-d991cbd4af09


build.gradle.kts
```diff
@@ -206,7 +212,14 @@ tasks {
...
         // Pass Segment write key for dev telemetry testing: ./gradlew runIde -Dconfluent.intellijplugin.segment.writeKey=your_key
-        System.getProperty("confluent.intellijplugin.segment.writeKey")?.let { systemProperty("confluent.intellijplugin.segment.writeKey", it) }
+        System.getProperty("confluent.intellijplugin.segment.writeKey")
+            ?.let { systemProperty("confluent.intellijplugin.segment.writeKey", it) }
+
+        // open a project directly, skipping the Welcome window
+        // (override with: IDEA_PROJECT_PATH=/path/to/project before running `gradle runIde`)
+        argumentProviders += CommandLineArgumentProvider {
+            listOf(System.getenv("IDEA_PROJECT_PATH") ?: "$projectDir")
+        }
```
shouples, line 222: This is the main addition; every other change in this file is due to autoformatting

.run/Run Plugin Clean.run.xml
shouples, line 1: Same as `.run/Run Plugin.run.xml`, just with `clean` (and a suffix added to the name)

Image

noeldevelops (approved):
formatting differences aside, this is working well locally. It'll be really nice to have that default project path to avoid a few clicks each time...

build.gradle.kts
```diff
@@ -206,7 +212,14 @@ tasks {
         System.getProperty("ccloud.callback-port")?.let { systemProperty("ccloud.callback-port", it) }
         System.getProperty("ccloud.env")?.let { systemProperty("ccloud.env", it) }
         // Pass Segment write key for dev telemetry testing: ./gradlew runIde -Dconfluent.intellijplugin.segment.writeKey=your_key
-        System.getProperty("confluent.intellijplugin.segment.writeKey")?.let { systemProperty("confluent.intellijplugin.segment.writeKey", it) }
+        System.getProperty("confluent.intellijplugin.segment.writeKey")
+            ?.let { systemProperty("confluent.intellijplugin.segment.writeKey", it) }
+
+        // open a project directly, skipping the Welcome window
+        // (override with: IDEA_PROJECT_PATH=/path/to/project before running `gradle runIde`)
```
flippingbits, line 219: @shouples Could we let it default to the current directory (`.`), so that `runIde` opens this project by default, eliminating the need for setting the env var?

shouples: As far as I understand, `$projectDir` should be pointing at the current `intellij` directory - or do you mean something else?

PR #358, opened 2026-02-09
Add Claude preference: use thisLogger() instead of println

Cerchie:
## Summary

Updates CLAUDE.md to clarify the existing logging guidance: use `thisLogger().debug()` instead of `println()` for debug output in the plugin code.

## Context

This preference was identified while replacing `println` statements with proper debug logging in `ScaffoldHttpClient.kt`. The CLAUDE.md file already mentioned using `thisLogger()` for logging, but this makes it explicit that `println` should not be used for debug output.

## Changes

- Enhanced the "Code Style" section to explicitly discourage `println` for debug output
- Clarified that `thisLogger().debug()` is the preferred approach


Cerchie:
context: Claude generated code with `println` everywhere, which is the equivalent of `console.log` so I figure we don't want that, used a custom slash command to generate this PR which will hopefully help clarify

shouples:
> context: Claude generated code with `println` everywhere, which is the equivalent of `console.log` so I figure we don't want that, used a custom slash command to generate this PR which will hopefully help clarify

Out of curiosity, is Claude able to access the logs easier with `println`? Or is it able to see the logs (in `idea.log`?) from `thisLogger`?

Cerchie:
> > context: Claude generated code with `println` everywhere, which is the equivalent of `console.log` so I figure we don't want that, used a custom slash command to generate this PR which will hopefully help clarify
> 
> Out of curiosity, is Claude able to access the logs easier with `println`? Or is it able to see the logs (in `idea.log`?) from `thisLogger`?

I'm not sure... I asked Claude, and the response implies it I guess? 

```
  If I had previously suggested println, it would have been a mistake - likely because println is simpler to write and might work for quick local debugging, but it's not appropriate for
  committed code in a production plugin.
```

flippingbits (approved):

PR #359, opened 2026-02-10
Add unit test coverage for native consumer

bcscc:
## Summary of Changes

- Add unit test coverage for the native/direct Kafka consumer pipeline in the IntelliJ plugin                                                        
- Tests covering filter logic, limit parsing, offset calculation, config serialization, and client lifecycle

Covered Classes:
- `ConsumerFilter` (37 tests): isRecordPassFilter() across all four filter types applied to key, value, header key, and header value fields. Validates AND semantics when multiple filter fields are set, null record field handling, non-string value coercion via toString(), and header byte array decoding.
- `ConsumerLimit` (11 tests): Computed property parsing (topicRecordsCount, topicRecordsSize, partitionRecordsCount, partitionRecordsSize) from type + value string pairs, including invalid/empty string edge cases and the DATE time field.
- `ConsumerEditorUtils` (18 tests): parsePartitionsText() range expansion (single values, comma-separated, hyphen ranges, mixed) and getStartWith() construction from UI inputs including offset negation for LATEST_OFFSET_MINUS_X, date-to-timestamp conversion, and blank/non-numeric offset handling.
- `KafkaOffsetUtils, KafkaOffsetUtilsCalculatedOffsetsTest` (20 tests): calculateStartTime() pure function for all ConsumerStartType values. calculateOffsets() suspend function covering offset-based types, timestamp-based types with KafkaDataManager.getOffsetsForData() mock, fallback to startOffset when offsets return -1, and error cases for out-of-range offsets and null start/end offsets.
- `StorageConsumerConfig` (12 tests: Round-trip serialization through the secondary constructor (which calls the private companion createFilterMap/createLimit/createStartWithMap methods) and getter deserialization for filter, limit, startWith, and field type accessors. Fallback behavior when persisted data contains unknown enum values (defaults to NONE/NOW/STRING).
- `KafkaConsumerClient` (5 tests): Constructor wiring, initial isRunning state, stop()/dispose() lifecycle, and callback invocation.

Known gaps
- `KafkaConsumerClient.start()/startInner()` is not unit tested. This is where the polling loop, partition assignment, seeking, limit enforcement, filter application, and error handling (RecordDeserializationException, SerializationException) all converge. It cannot be tested without a live Kafka broker because createConsumer() instantiates a KafkaConsumer internally with no injection point for a mock.

1.

-

## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


test/io/confluent/intellijplugin/util/KafkaOffsetUtilsTest.kt
```diff
@@ -0,0 +1,132 @@
...
+        fun `should return timestamp approximately 1 hour ago for LAST_HOUR`() {
+            val startWith = ConsumerStartWith(
+                type = ConsumerStartType.LAST_HOUR, time = null, offset = null, consumerGroup = null
+            )
+            val result = KafkaOffsetUtils.calculateStartTime(startWith)
+            assertNotNull(result)
+            val expectedTime = System.currentTimeMillis() - 3600000L
+            val tolerance = 2000L
+            assertTrue(
+                kotlin.math.abs(result!! - expectedTime) <= tolerance,
```
flippingbits, line 38: This might cause the test to become flaky. If it becomes an issue, we could consider mocking the time/clock.

flippingbits (approved):
LGTM, thank you! I have one minor comment but it should not block the PR.

PR #357, opened 2026-02-09
Template selection: HTTP client code 

Cerchie:
## Summary of Changes

> [!NOTE]
> I previously had the action registered and the UI changes created and tested, but that resulted in a LOT to review, so limiting it to the client for now. It's still a lot, but this is due to the patch now as well as thorough unit tests. 

Partially addresses #351 

This PR looks large due to the patch required to remove `any` types (Kotlin seems to treat the spec object differently than TS), I recommend collapsing the gen/ and openapi/ folders to review the business logic first. See VS Code for a similar[ /patches](https://github.com/confluentinc/vscode/tree/main/src/clients/sidecar-openapi-specs/patches) folder.
Screenshot 2026-02-18 at 7 14 07 AM

The PR adds a `fetchTemplates` [function](https://github.com/confluentinc/intellij/pull/357/changes#diff-e74a69fe979158eae153565740fdcd580d9598c67c02eb2a369ac16bf8a27b77R82) to provide templates which can then be displayed using Actions and the UI DSL, in incoming branches. I also patched the openapi spec due to a bunch of `any` types I found in there while working on this. 

**Current line test coverage at 84% for the client, SQ has it at 61%** but that includes models so I'm going with this one:
Screenshot 2026-02-17 at 11 15 58 AM

### Optional: Click-testing instructions

For running only the scaffold tests:  `./gradlew test --tests "*ScaffoldHttpClientTest"`

### Optional: Any additional details or context that should be provided?

Next branch up will include UI, but this branch needs to be merged first. 

## Pull request checklist

Please check if your PR fulfills the following (if applicable):

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing

PR #372, opened 2026-02-17
update CLAUDE.md testing code instructions

Cerchie:
## Summary of Changes

Ran into this while I was working on https://github.com/confluentinc/intellij/pull/357 

Hopefully now Claude will "know" to add fixtures rather than inline json for readability as well as editability. 

Prompt was: "❯ Edit your md so you don't use inline json strings in tests, instead, you create separate fixture"


CLAUDE.md
```diff
@@ -152,6 +152,39 @@ test {
...
+    // ...
+}
+```
+
+**Good:**
+
+```kotlin
+@Test
+fun `should parse response`() {
+    val json = javaClass.getResourceAsStream("/fixtures/sample-response.json")!!.readText()
```
shouples, line 175: Are .json files the preferred way of handling test helper data/fixtures in Kotlin? Or should this use a factory function / singleton pattern like `vscode` for its test resources?

Cerchie: oh good question, I was going off our standards for the sidecar (see [example](https://github.com/confluentinc/ide-sidecar/tree/main/src/test/resources/kafka-rest-proxy-mock-responses)) but that's Java. @flippingbits , do you have an opinion here?

Cerchie: oh & see some precedent here https://github.com/confluentinc/intellij/tree/main/test/resources/ccloud-resources-mock-responses

flippingbits: Good question! I believe this is the response of a mocked CCloud API endpoint, in which case JSON makes a lot of sense. @Cerchie, maybe you could clarify it in the `CLAUDE.md`?

Cerchie: yes, I added a phrase to narrow it down to mocked API endpoints

flippingbits (approved):

PR #373, opened 2026-02-17
Add unit test coverage for consumer module

bcscc:
### Summary of Changes

- Add `ConsumerEditorUtilsTest` export function tests (JSON, CSV, TSV) covering `getTableContent`, null handling, quote escaping, whitespace normalization
- Create `KafkaRecordTest` with tests for `createFor` factory methods (success/failure paths), computed properties (`keyText`, `valueText`, `errorText`), and default value handling
- Create `ConsumerProducerFieldConfigTest` covering `getValueObj()` type conversion logic for all primitive types (STRING, JSON, LONG, INTEGER, DOUBLE, FLOAT, BASE64, NULL) and error cases
- Add edge case tests to `ConsumerFilterTest` for `DOES_NOT_CONTAINS` and `REGEX` with null record key/value, and empty headers scenarios
- Fix KafkaRecord.createFor failure path using keyFormat for both keyFormat and valueFormat instead of using the valueFormat parameter

N/A — unit tests only.

- `KafkaConsumerClient` private methods (`calculatePartitions`, `partitionOffsetsForStartOffset`, `partitionOffsetsForStartDate`) were evaluated for testability but skipped — they're not reachable through the public API without mocking the full `KafkaConsumer` chain
- The JSON export null-handling bug discovered during testing is addressed in a separate follow-up PR

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing


test/io/confluent/intellijplugin/consumer/editor/ConsumerEditorUtilsTest.kt
```diff
@@ -158,4 +167,145 @@ class ConsumerEditorUtilsTest {
...
+        fun `should return empty JSON array for empty table`() {
+            val model = createTableModel(listOf("Col1", "Col2"), emptyList())
+            val result = ConsumerEditorUtils.getTableContent(model, "json")
+            assertEquals("[]", result.trim())
+        }
+
+        @Test
+        fun `should produce valid JSON with column names as keys for single row`() {
+            val model = createTableModel(listOf("Name", "Age"), listOf(listOf("Alice", "30")))
+            val result = ConsumerEditorUtils.getTableContent(model, "json")
```
flippingbits, line 215: nit: Can we check that `result` holds valid JSON?

src/io/confluent/intellijplugin/consumer/editor/KafkaRecord.kt
```diff
@@ -80,7 +80,7 @@ data class KafkaRecord(
                     valueSize = 0,
                     headers = emptyList(),
                     keyFormat = keyFormat ?: KafkaRegistryFormat.UNKNOWN,
-                    valueFormat = keyFormat ?: KafkaRegistryFormat.UNKNOWN
+                    valueFormat = valueFormat ?: KafkaRegistryFormat.UNKNOWN
```
flippingbits, line 83: Nice catch! Could we cover this with a test that assigns different values to the `keyFormat` and `valueFormat`?

flippingbits (dismissed):
LGTM!

SonarQube reported 50+ issues but most (if not all?) of them seem to be related to the naming of the test cases. I feel we can ignore it.

src/io/confluent/intellijplugin/consumer/editor/KafkaRecord.kt
bcscc: Already added [here](https://github.com/confluentinc/intellij/pull/373/changes#diff-5a104b54685db3f3506a9927dd00287d1e3f90b9d60d9a0382f0152c9b95b012R206-R215)!

test/io/confluent/intellijplugin/consumer/editor/ConsumerEditorUtilsTest.kt
bcscc: Thanks, added!

flippingbits (approved):
LGTM 🎉

PR #374, opened 2026-02-17
Fix null cell NPE in JSON table export

bcscc:
### Summary of Changes

- Fix `NullPointerException` in `ConsumerEditorUtils.getJTableAsJson` when a table cell value is null
  - `JsonPrimitive(null)` throws NPE — replaced with `JsonNull.INSTANCE` for null cells
  - Added `.serializeNulls()` to `GsonBuilder` so Gson actually includes null entries in output (without this, null values are silently dropped producing `{}` instead of `{"Col": null}`)
- Add test verifying null cells export as proper JSON `null` values

1. Open a consumer, consume records that have null keys or values
2. Export table as JSON
3. Verify null fields appear as `null` in the JSON output instead of crashing or being omitted

- Discovered while writing export tests in #373
- The CSV export path already handled nulls correctly (renders as `""`) — only the JSON path was affected
- `KafkaEditorUtils.tryFormatJson` already uses `.serializeNulls()` — this brings `getJTableAsJson` in line with that pattern

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing


flippingbits (approved):
LGTM, good catch!

PR #392, opened 2026-02-19
Fix memory leak in BdtMonitoringUpdater disposable registration

noeldevelops:
#   🤖 Generated with Claude Code
## Summary

Fixes a memory leak in `BdtMonitoringUpdater` where a disposable was created without a proper parent, causing it to be registered under `ROOT_DISPOSABLE`. This
  resulted in SEVERE error logs on IDE shutdown when the disposable wasn't properly cleaned up.

  **Change:** Modified line 187 in `BdtMonitoringUpdater.kt` to register the disposable with `this` as the parent instead of leaving it parentless.

```kotlin
  // Before
  val disposable = Disposer.newDisposable()

  // After
  val disposable = Disposer.newDisposable(this)
```

### Problem

When a disposable is created without a parent using Disposer.newDisposable(), it gets registered as a child of ROOT_DISPOSABLE. If the isCurrentCanceled.afterChange callback never fires (or fires inconsistently), the disposable is never explicitly disposed, causing a memory leak that the IntelliJ Platform detects on shutdown.

### Reproduction

  1. Build and run the plugin: ./gradlew runIde
  2. Open the Kafka tool window (bottom panel)
  3. Add a new Kafka cluster connection or use an existing one
  4. Connect to the cluster and browse topics/consumer groups (this triggers BdtMonitoringUpdater.reloadAll() which creates the leaked disposable)
  5. Close the IDE completely
  6. Check the IDE logs (~/Library/Logs/JetBrains/IntelliJIdea*/idea.log on macOS)
  7. Look for SEVERE errors from com.intellij.openapi.util.ObjectTree mentioning:
  Memory leak detected: 'newDisposable' (class com.intellij.openapi.util.Disposer$1)
  was registered in Disposer as a child of 'ROOT_DISPOSABLE' but wasn't disposed.

  The stack trace will point to BdtMonitoringUpdater.kt:198 (line 187 in the source code).

### Solution
By passing this to Disposer.newDisposable(this), the disposable is now registered as a child of BdtMonitoringUpdater. This ensures it will be automatically cleaned up when the updater itself is disposed, even if the callback never fires.

This follows the IntelliJ Platform best practice documented at https://plugins.jetbrains.com/docs/intellij/disposers.html

### Test Plan

  - Run the plugin with the fix
  - Connect to a Kafka cluster and browse topics
  - Close the IDE
  - Verify no SEVERE memory leak errors appear in IDE logs


Cerchie (approved):
this looks good to me. I grepped for the error on `main`, then cleared the log files and explored topics on your branch, didn't find it on your branch. 

Image
Image

PR #393, opened 2026-02-19
Consumer - new message viewer experience dev flag

### LINKED ISSUE #380: [Consumer] Add feature or env flag for new message viewer
To get started we at least want a developer-facing flag to enable/disable the new Message Viewer UI while we iterate on it. TBD best DX to implement - a env/debug flag or whatever. Note: User-facing options for hiding/flagging the upgrade are currently under discussion with team, this is for devs only.

noeldevelops:
## Summary of Changes

- Adds a new system property that can be passed during build to enable the new message viewer experience
- Closes #380. A first step toward the Message Viewer upgrade work.
- ⚠️ Flipping the flag to `true` won't do much, since we still need to implement the actual browser view

1. Run the plugin build with the property set to `true`
``` bash
./gradlew runIde -Dconfluent.intellijplugin.consumer.webview=true
```
2. When you open the Consumer interface for any topic, you should see an Info log print to confirm the flag was set. It says "_WebView message viewer flag enabled, but implementation not ready. Using default viewer._"

- Not user-facing! 
- Open to an `env` var instead if that is preferable for the team? See the [first commit](https://github.com/confluentinc/intellij/commit/6a02b31b626ac68d549933e80bbbdf9ee0b0e9ed) where Claude did both.


src/io/confluent/intellijplugin/consumer/editor/KafkaConsumerPanel.kt
```diff
@@ -56,7 +57,28 @@ class KafkaConsumerPanel(
         onStart = ::onStartConsume,
         onStop = ::onStopConsume
     )
-    private val output = KafkaRecordsOutput(project, isProducer = false).also { Disposer.register(this, it) }
+
+    // Feature flag to enable new message viewer UI (WebView-based)
+    // Enable via: -Dkafka.message.viewer.webview=true
+    // Default: false (uses existing Swing-based table viewer)
+    private val useWebViewMessageViewer: Boolean =
+        System.getProperty("kafka.message.viewer.webview")?.toBoolean() ?: false
```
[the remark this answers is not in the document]
noeldevelops: oh good catch I need to update this

src/io/confluent/intellijplugin/consumer/editor/KafkaConsumerPanel.kt
```diff
@@ -56,7 +57,31 @@ class KafkaConsumerPanel(
...
-    private val output = KafkaRecordsOutput(project, isProducer = false).also { Disposer.register(this, it) }
+
+    // Feature flag to enable new message viewer UI (WebView-based)
+    // Default: false (uses existing Swing-based table viewer)
+    private val useWebViewMessageViewer: Boolean =
+        System.getProperty(CONSUMER_WEBVIEW_PROPERTY)?.trim()?.takeIf { it.isNotBlank() }?.let { value ->
+            value.toBooleanStrictOrNull() ?: false.also {
+                thisLogger().warn("Invalid value for $CONSUMER_WEBVIEW_PROPERTY: '$value'. Expected 'true' or 'false'. Defaulting to false.")
+            }
+        } ?: false
```
flippingbits, line 68: nit: We'd be able to easily test (and reuse) the parsing functionality, if we moved it to a separate function.

flippingbits (approved):
Thanks, @noeldevelops! Will this be a long-term feature flag or do we plan to deprecate it once the webview-based message viewer is in a stable state?

noeldevelops:
@flippingbits we plan to deprecate the flag once the feature is stable

src/io/confluent/intellijplugin/consumer/editor/KafkaConsumerPanel.kt
noeldevelops: Good idea. This is similar to the way we parse other system flags and probably is a good thing to extract, esp if we continue to use this pattern. I will follow up.

PR #396, opened 2026-02-20
Exclude generated source code from test cov report

flippingbits:
## Summary of Changes

This change excludes the auto-generated source code, stored in `/gen`, from the test coverage report.

1.

-

## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


sonar-project.properties
```diff
@@ -1,10 +1,10 @@
 ### service-bot sonarqube plugin managed file
-sonar.coverage.exclusions=**/test/**/*,**/tests/**/*,**/mock/**/*,**/mocks/**/*,**/*mock*,**/*test*
+sonar.coverage.exclusions=**/test/**/*,**/tests/**/*,**/mock/**/*,**/mocks/**/*,**/*mock*,**/*test*,gen/**/*
 sonar.coverage.jacoco.xmlReportPaths=**/jacocoTestReport.xml
 sonar.cpd.exclusions=**/test/**/*,**/tests/**/*,**/mock/**/*,**/mocks/**/*,**/*mock*,**/*test*
```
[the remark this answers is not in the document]
flippingbits: Unfortunately, we can't set this via ServiceBot.

service.yml
```diff
@@ -55,4 +55,14 @@ sonarqube:
...
+  coverage_sources:
+    - "src"
+  coverage_exclusions:
+    - "**/test/**/*"
+    - "**/tests/**/*"
+    - "**/mock/**/*"
+    - "**/mocks/**/*"
+    - "**/*mock*"
+    - "**/*test*"
+    - "gen/**/*"
```
shouples: Minor: we shouldn't need this if `coverage_sources` starts at the `src/` level
```suggestion
```

service.yml
```diff
@@ -55,4 +55,14 @@ sonarqube:
...
     - "**/jacocoTestReport.xml"
+  coverage_sources:
+    - "src"
+  coverage_exclusions:
+    - "**/test/**/*"
+    - "**/tests/**/*"
+    - "**/mock/**/*"
+    - "**/mocks/**/*"
+    - "**/*mock*"
+    - "**/*test*"
```
shouples: Does any of this exist under `src/`? Might be able to remove the whole section

shouples (dismissed):
Setting the `coverage_source` seems like it will help the most here, so I'm not sure if `coverage_exclusions` are necessary based  on the rest either being under `gen/` or `test/` but also won't hurt

flippingbits:
@shouples Fair point! Let's try it out without configuring `coverage_exclusions` and add it back if needed.

shouples (approved):

PR #421, opened 2026-03-02
Schema Registry deserialization and type selection for CCloud consumer

### LINKED ISSUE #405: Schema Registry Deserialization for CCloud Consumer
`CCloudConsumerClient.extractValue()` receives `__raw__` base64 bytes but does no schema-aware decoding. Avro/Protobuf/JSON Schema messages display as raw binary.

After base64-decoding `__raw__`, look up the schema ID from the wire-format bytes (first 5 bytes = magic byte + 4-byte schema ID), fetch the schema from CCloud SR via `DataPlaneFetcher`, and deserialize using existing
`AvroSchemaUtils/ProtobufSchemaUtils/JsonSchemaUtils`.

### LINKED ISSUE #406: Enable Key/Value Type Selection for CCloud Consumer
The consumer panel's format dropdowns is currently gated by `supportsSchemaRegistry()` which returns false for CCloud. Enable the key/value type selector in `KafkaConsumerPanel` for CCloud. For primitive types (STRING, JSON, LONG, etc.), apply client-side conversion to the REST response data while for SCHEMA_REGISTRY populate based on the fetched schema SR.

bcscc:
### Summary of Changes

- Add schema-aware deserialization to `CCloudConsumerClient` for Avro, Protobuf, and JSON Schema messages (reference [PR from ide-sidecar](https://github.com/confluentinc/ide-sidecar/pull/525))
  - V0 wire format: detect magic byte `0x00` + 4-byte schema ID prefix in payload, fetch schema by ID from CCloud SR via `DataPlaneFetcher`, then deserialize using the appropriate format-specific decoder
  - V1 wire format: extract schema GUID from `confluent.key.schemaId` / `confluent.value.schemaId` headers (magic byte `0x01` + 16-byte UUID), fetch schema by GUID via new `getSchemaByGuid()` endpoint
  - Schema cache (`ConcurrentHashMap`) keyed by schema ID or GUID to avoid redundant SR fetches within a session
  - Deserialization errors on individual records are reported via `consumeError()` without stopping the consumer
- Use `return_raw_base64_records=true` query parameter on the consume endpoint
   - Forces CCloud REST API to return ALL records as `{"__raw__": "<base64>"}` format, providing consistent binary access regardless of content type
   - Records are decoded once via `decodeRawBytes()` into `ByteArray`, then passed through the entire pipeline — eliminates redundant JSON-to-type conversion and enables accurate byte-level sizing
   - Removes dead code paths: `estimateJsonSize()`, `convertJsonToType()`, and the plain `JsonElement` branch in `extractValue()`
- Enable key/value type-aware conversion in `CCloudConsumerClient`
  - Primitive types (STRING, LONG, INTEGER, DOUBLE, FLOAT, BASE64, JSON, NULL) apply client-side conversion using cached Kafka `Deserializer` instances created at `start()` time
  - SCHEMA_REGISTRY type routes through the full schema-aware deserialization pipeline
  - Plain JSON values (non-binary) are converted to the selected type via `convertJsonToType()`
- Complete schema registry UI layer methods in `CCloudClusterDataManager`
  - `listSchemasNames()`, `updateSchemaList()`, `getSchemasForEditor()`, `getLatestVersionInfo()`, `getCachedOrLoadSchema()` now delegate to `DataPlaneFetcher` instead of returning empty/unsupported
- Add `KafkaRegistryFormat.fromSchemaType()` to map SR schema type strings to the internal format enum (null defaults to AVRO per SR convention)
- Add `DataPlaneFetcher.getSchemaByGuid()` and corresponding `CloudConfig.SCHEMA_BY_GUID_URI` for V1 header-based schema lookups
- Fix header decoding: CCloud REST API returns header values as base64-encoded strings, now properly decoded to `byte[]` so schema GUIDs in headers can be detected
- Remove unused `supportsSchemaRegistry()` method from `BaseClusterDataManager` — the UI already gates type selector visibility via `registryType` directly in `KafkaEditorUtils.createFieldTypeComboBox()`

Fixes #405
Fixes #406 

1. Open a CCloud-connected topic consumer
2. Verify key/value type dropdowns are enabled (STRING, LONG, JSON, SCHEMA_REGISTRY, etc.)
3. Consume messages from a topic with Avro-encoded values — verify they render as readable JSON instead of raw binary
4. Repeat with Protobuf and JSON Schema encoded topics
5. Test primitive type selection: switch value type to LONG/INTEGER for topics with numeric keys/values
6. Test with both V0 (schema ID in payload) and V1 (schema GUID in headers) wire formats
7. Select SCHEMA_REGISTRY — verify schema list populates in the dropdown

- Before: `extractValue()` returned raw base64 bytes for schema-encoded messages; type selector dropdowns were gated by `supportsSchemaRegistry()` (unused dead method); schema registry methods in `CCloudClusterDataManager` were stubbed out
- After: Messages are deserialized according to their schema format and displayed as readable text; all primitive and schema types are selectable; schema list, versioning, and editor integration work for CCloud
- Schema deserialization reuses existing `KafkaRegistryUtil.parseSchema()` and format-specific utilities (`AvroSchemaUtils`, `ProtobufSchemaUtils`) already used by the native consumer path

## Pull request checklist

#### Tests

- [X] Added new
- [ ] Updated existing
- [ ] Deleted existing


src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
```diff
@@ -59,7 +77,20 @@ class CCloudConsumerClient(
     private var consumerScope: CoroutineScope? = null
 
     // Track next offsets per partition for subsequent requests
-    private val nextOffsets = mutableMapOf<Int, Long>()
+    @VisibleForTesting
+    internal val nextOffsets = mutableMapOf<Int, Long>()
+
+    // Cache parsed schemas by schema ID or GUID string to avoid redundant fetches
+    private val schemaCache = ConcurrentHashMap<String, ParsedSchema>()
```
flippingbits: Can we scope the cache by the SR cluster ID? IIRC, different schema registries can hold schemas with the same ID.

src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
```diff
@@ -428,38 +478,225 @@ class CCloudConsumerClient(
     }
 
     /**
-     * Extract value from JSON element.
-     * Handles both plain values and schema-encoded values ({"__raw__": "base64"}).
+     * Extract value from JSON element with type-aware conversion.
+     * - SCHEMA_REGISTRY → full schema-aware deserialization (Avro/Protobuf/JSON Schema)
+     * - Other types → decode raw bytes and convert to the selected primitive type
```
flippingbits: We can force the CCloud consume endpoint to always return raw bytes by adding the query param `return_raw_base64_records=true`. I think it makes sense to use it here unless we already do so. I'd be happy to sync on this with you.

src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
```diff
@@ -130,13 +166,17 @@ class CCloudConsumerClient(
 
                 timestampUpdate()
 
-                // Update next offsets from response
                 updateNextOffsets(response)
 
-                // Flatten all records from all partitions
+                // Flatten all records from all partitions, skipping records with deserialization errors
```
noeldevelops: what happens to the skipped records? is there an easy way to click test consume with a deserialization error?

src/io/confluent/intellijplugin/registry/KafkaRegistryFormat.kt
```diff
@@ -11,5 +11,15 @@ enum class KafkaRegistryFormat(@Nls val presentable: String) {
...
+
+        /**
+         * Map Schema Registry schema type string to format enum.
+         * Per SR convention, null schemaType defaults to AVRO.
+         */
+        fun fromSchemaType(type: String?): KafkaRegistryFormat = when (type?.uppercase()) {
+            "PROTOBUF" -> PROTOBUF
+            "JSON" -> JSON
+            else -> AVRO  // null or "AVRO"
+        }
```
flippingbits, line 25: nit: Could we handle the case where `type` is an unexpected value by returning `UNKNOWN` (defined in line 10)?

flippingbits (commented):
Nice work! I have a few minor questions. Otherwise, your PR looks good to me.

src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
bcscc: Sorry that comment is a bit misleading, records with deserialization failures are reported as error rows in the consumer output via consumeError() and excluded from the batch (allRecords list). The user sees the error message along with the partition and offset. Consumption continues for remaining records. I've updated the comment.

A easy way to click-test is consume from a topic with Schema-encoded values while the value type is set to something like a primitive type like LONG, the byte deserialization will fail and you'll see error rows appear in the output.

src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
bcscc: Done, cache keys are now prefixed with the SR cluster ID!

src/io/confluent/intellijplugin/registry/KafkaRegistryFormat.kt
bcscc: Done, unrecognized type strings now return UNKNOWN!

src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
bcscc: That does make it a lot simpler with the query param. I've made the [change](https://github.com/confluentinc/intellij/pull/421/changes/e0365349be1de6f70fc38f0768fd1c9a3dce6fc9) to default consume from that endpoint option, please let me know if there's anything else I can improve on! Thanks!

src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
```diff
@@ -59,7 +76,21 @@ class CCloudConsumerClient(
     private var consumerScope: CoroutineScope? = null
 
     // Track next offsets per partition for subsequent requests
-    private val nextOffsets = mutableMapOf<Int, Long>()
+    @VisibleForTesting
+    internal val nextOffsets = mutableMapOf<Int, Long>()
+
+    // Cache parsed schemas by schema ID or GUID string to avoid redundant fetches
+    @VisibleForTesting
+    internal val schemaCache = ConcurrentHashMap<String, ParsedSchema>()
```
flippingbits, line 84: nit:

If we model Schema Registry IDs and Schema IDs/GUIDs as value classes:

```kotlin
@JvmInline
internal value class SchemaRegistryClusterId(val id: String)

internal sealed interface SchemaCacheKey {
    @JvmInline
    value class ById(val schemaId: Int) : SchemaCacheKey

    @JvmInline
    value class ByGuid(val guid: UUID) : SchemaCacheKey
}
```

We could make the cache type safe and won't have to rely on the caller to come up with the right cache key:
```kotlin
internal val schemaCache = ConcurrentHashMap<SchemaRegistryClusterId, ConcurrentHashMap<SchemaCacheKey, ParsedSchema>>()
```

flippingbits (approved):
Thanks for the update! I have one more minor comment but it shouldn't block the PR.

src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
bcscc: Thanks! Will address this is a follow up PR

PR #430, opened 2026-03-04
Enable preset and details panel for CCloud consumer

### LINKED ISSUE #407: Enable Consumer Details Panel for CCloud Consumer
`supportsDetailsPanel()` currently returns false for CCloud. Users can't inspect individual record details (headers, metadata, key/value in different views like JSON/BASE64). Enable this panel for CCloud consumer and ensure `KafkaRecordDetails` renders correctly for CCloud records (headers from REST response, size info, type labels)

### LINKED ISSUE #408: Enable Consumer Presets for CCloud Consumer
CCloud consumer currently has `supportsPresets()` set to false, can't save/load consumer configurations, must reconfigure every session. Enable this feature and verify `StorageConsumerConfig` persistence works for CCloud consumer settings, enable preset panel in `KafkaConsumerPanel`.

bcscc:
### Summary of Changes

- Enable details panel and preset panel for CCloud OAuth connection consumer
  - Both panels were previously gated behind `supportsPresets()` / `supportsDetailsPanel()` capability flags that returned `false` for CCloud; now unconditionally shown for all consumer types
  - Save preset action (bookmark icon) is also now available on CCloud consumer panels
- Scope saved presets by connection type so CCloud and native presets are kept separate
  - Added `connectionType: String?` field to `StorageConsumerConfig` (defaults to `null` for backwards compatibility — existing native presets continue to work without migration)
  - Native connections save presets with `connectionType = null`; CCloud connections tag presets with `connectionType = "ccloud"`
  - `KafkaRunConfig` extended with an optional filter predicate; `addConfig` / `removeConfig` operate on the full unfiltered storage to avoid corrupting presets from other connection types
  - `KafkaConfigStorage.consumerConfigFor(connectionType)` returns a scoped `KafkaRunConfig` view that shares underlying storage and change listeners with all other views
  - `Presets.configAdded()` checks `runConfig.shouldShow()` before adding to the list model so newly saved presets from other connection types don't appear in the wrong panel
  - `BaseClusterDataManager.presetConnectionTag()` returns `null` by default; `CCloudClusterDataManager` overrides to return `"ccloud"`

**Merge #421 before this PR**

Fix #407 
Fix #408 

1. Open a CCloud-connected topic consumer — verify the Presets and Details panels are visible
2. Save a preset from the CCloud consumer, then open a native consumer — verify the CCloud preset does not appear
3. Save a preset from a native consumer, then open a CCloud consumer — verify the native preset does not appear
4. Verify existing native presets (saved before this change) still appear on native consumers

- Before: CCloud consumer had no preset panel, no details panel, and no save preset action; presets were shared in a flat global list across all connection types
- After: CCloud consumer has full preset and details panel support; presets are siloed by connection type with `null` for native and `"ccloud"` for CCloud

## Pull request checklist

#### Tests

- [X] Added new
- [ ] Updated existing
- [ ] Deleted existing


noeldevelops (approved):
Testing locally and opening/closing details, adding favorites (presets) + restoring from them, etc all working as expected. Code to filter by type looks good as well. ✅

PR #431, opened 2026-03-04
Consumer performance improvements: batch table updates, use ArrayDequeue for storage

### LINKED ISSUE #447: Get rid of ArrayList.removeFirst(), an O(n^2) operation
Change list data type to avoid O(n) table updates after reaching max message count. Explore `ArrayDequeue` 

I suspect this is the culprit in most message viewer freezes at high consumer volumes

noeldevelops:
## Summary of Changes
 
- Closes https://github.com/confluentinc/intellij/issues/447
- Replace `LinkedList`  in `KafkaRecordsOutput` with a pre-sized `ArrayDequeue` for O(1) random access and better memory usage
- Change to batched table updates in `ListTableModel` — accumulates pending items and flushes them in a single EDT event via `invokeLater` instead of for every single row (reduces UI events in this table by 100-500x during streaming)
- Update tests to cover new batching, capacity enforcement, etc

To test performance improvements and verify basic actions still work as expected, try interacting with the Consumer  UI while consuming a high volume/throughput. 

For example:
#### Consumer
- [x] Open a topic with a moderate number of messages (>100). Consume and verify all rows appear in the table with correct Topic, Timestamp, Key, Value, Partition, and Offset columns
- [x] While consuming, esp for a high-throughput topic (1K+ msgs/sec), confirm the UI stays responsive — you should be able to scroll, click rows, resize columns, and expand/collapse panels without freezing
- [x] Use column header filters (text search) to filter by key or value content. Confirm filtering works and results update
- [x] Click the clear button (trash icon) while consuming. Confirm the table empties and new messages continue to appear
- [x] Set a limit in `advanced settings -> consumer max limit` to something low you can see in the table (10 or 20) and consume beyond it. Confirm oldest rows are evicted and the table doesn't grow beyond the limit
- [x] Sort by any column (Timestamp, Partition, etc.) during active consumption. Confirm sorting still works correctly with batched updates

- Try the interactions above on `main` or current plugin for a performance comparison with your own eyes !
- The Producer also benefits from the `ArrayDeque` switch; no Producer-specific changes were needed because `KafkaRecordsOutput` is the shared component.

## Pull request checklist

#### Tests

- [X] Added new
- [X] Updated existing
- [x] Deleted existing


noeldevelops:
[❗ Claude wrote this]
Addressed both review comments on the `ArrayList` eviction cost:

Switched the backing store to `ArrayDeque` (circular buffer) which gives O(1) head removal while preserving O(1) random access. The eviction now uses `repeat(n) { data.removeFirst() }` which is O(n) in elements removed, not O(data.size). 

The `subList(0, n).clear()` suggestion would still be O(data.size) with `ArrayList` due to element shifting, so switching the data structure was the cleaner fix.

See 2ce5586a.

src/io/confluent/intellijplugin/common/editor/ListTableModel.kt
```diff
@@ -51,20 +56,71 @@ class ListTableModel<T>(
...
+    private fun scheduleFlush() {
+        if (!flushScheduled) {
+            flushScheduled = true
+            invokeLater {
+                try {
+                    flushPendingAdds()
+                } finally {
+                    flushScheduled = false
+                }
             }
```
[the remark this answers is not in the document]
noeldevelops: handled in finally when we re-check

src/io/confluent/intellijplugin/consumer/editor/KafkaRecordsOutput.kt
```diff
@@ -39,7 +39,7 @@ class KafkaRecordsOutput(val project: Project, val isProducer: Boolean) : Dispos
     private var tableLoadingDecorator: TableLoadingDecorator? = null
 
     internal val outputModel = ListTableModel(
-        LinkedList<KafkaRecord>(),
+        ArrayList<KafkaRecord>(1000),  // Pre-size for efficiency, O(1) random access
         listOf(TOPIC_FIELD, TIMESTAMP_FIELD, KEY_COLUMN, VALUE_COLUMN, PARTITION_COLUMN) +
                 if (isProducer) listOf(DURATION_COLUMN) else listOf(OFFSET_COLUMN)
```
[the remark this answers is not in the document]
noeldevelops: changed to circular buffer

src/io/confluent/intellijplugin/common/editor/ListTableModel.kt
```diff
@@ -51,21 +57,94 @@ class ListTableModel<T>(
...
+     * Synchronously replace all data. Unlike addBatch(), this does not defer via invokeLater,
+     * so callers can read elements() immediately after this call returns.
+     */
+    fun replaceAll(elements: List<T>) {
+        synchronized(pendingAdds) {
+            pendingAdds.clear()
+        }
+        flushScheduled.set(false)
+        data.clear()
+        data.addAll(elements)
```
Cerchie: L78 is the only different between this and the `clear()` fun above... I guess one is for adding incoming els and the other is for clearing the whole batch?

src/io/confluent/intellijplugin/common/editor/ListTableModel.kt
```diff
@@ -51,21 +57,94 @@ class ListTableModel<T>(
...
+            pendingAdds.addAll(elements)
+        }
+        scheduleFlush()
+    }
+
+    /**
+     * Add a single element. For backward compatibility.
+     * For better performance when adding multiple elements, use addBatch() instead.
+     */
+    fun addElement(element: T) = addBatch(listOf(element))
```
Cerchie: Since this just calls `addBatch`, do we need it? I see a certain advantage in readability but we could also maybe add a docstring to `addBatch` that makes it clear you can add one element

test/io/confluent/intellijplugin/common/editor/ListTableModelTest.kt
```diff
@@ -1,145 +1,314 @@
...
+        @Test
+        fun `clear should reset batching state so new adds flush correctly`() {
+            // Given - simulate the bug scenario: add data, clear, then add more
+            model.addBatch(listOf("old1", "old2"))
+            ApplicationManager.getApplication().invokeAndWait { }
+            events.clear()
+
+            // When - clear and then add new data
+            model.clear()
+            events.clear()
```
Cerchie: why do we need `events.clear()` once more after `model.clear()`?

src/io/confluent/intellijplugin/common/editor/ListTableModel.kt
noeldevelops: going to double-check this one

src/io/confluent/intellijplugin/common/editor/ListTableModel.kt
noeldevelops: agreed. Claude wrote this to be backward compatible, but we probably don't need to worry about that. I'll double check no other callers need this & delete

test/io/confluent/intellijplugin/common/editor/ListTableModelTest.kt
noeldevelops: good catch, think it's a duplicate.

src/io/confluent/intellijplugin/common/editor/ListTableModel.kt
noeldevelops: Ok so these are used in different places/ways, but the shared logic can be abstracted out. Doing it now !

src/io/confluent/intellijplugin/common/editor/ListTableModel.kt
noeldevelops: removing 🧼

test/io/confluent/intellijplugin/common/editor/ListTableModelTest.kt
noeldevelops: I'm removing it. Claude's explanation had an interesting reason for it, though: "The events.clear() after model.clear() was clearing the table-changed event fired by clear(), but the test never asserts on event counts, so it was unnecessary."

bcscc (approved):
LGTM, working a lot faster, thank you!

src/io/confluent/intellijplugin/consumer/editor/KafkaRecordsOutput.kt
bcscc, line 207:
```suggestion
        return outputModel.elements()
```
nit: `elements()` already returns a list so this is a slightly unnecessary convertion that causes an additional copy.

Cerchie (approved):

PR #434, opened 2026-03-05
Enable advanced consumer settings for CCloud connections

### LINKED ISSUE #409: Advanced Consumer Settings for CCloud Consumer
Currently `supportsAdvancedSettings()` is set to false in the `KafkaConsumerPanel`. CCloud users can't configure poll size, display limits, or other consumer parameters. Enable a subset of advanced settings that make sense for REST-based consumption (ex: max poll records, display last N records). Note: some native Kafka settings like don't apply to REST.

bcscc:
### Summary of Changes

- Make `KafkaConsumerSettings` connection-type-aware
  - Accepts a `supportedProperties` parameter to filter which Kafka consumer config properties are shown in the settings panel
  - CCloud connections show only `max.poll.records` and `fetch.max.bytes` (the two settings supported by the CCloud REST consume API)
  - Native connections continue to show all five properties (`request.timeout.ms`, `max.poll.records`, `fetch.max.wait.ms`, `fetch.max.bytes`, `max.partition.fetch.bytes`)
  - `KafkaConsumerPanel` checks if the manager is a `CCloudClusterDataManager` to select the appropriate property set
- Wire advanced settings into `CCloudConsumerClient` consume requests
  - `resolvedMaxPollRecords` and `resolvedFetchMaxBytes` are resolved from config at `start()` time
  - All `buildInitialConsumeRequest()` and `buildSubsequentConsumeRequest()` calls now pass `maxPollRecords` and `fetchMaxBytes` to the REST API
  - When unset, values are `null` — letting the server use its defaults (500 for `max_poll_records`, 50MB for `fetch_max_bytes`)
  - Values are reset on `stop()`
- Remove `supportsAdvancedSettings()` feature flag
  - Advanced settings are now always shown for all connection types
  - The `supportedProperties` filtering handles what's relevant per connection type, making the boolean flag unnecessary
- Hide empty properties section in settings panel when no Kafka config properties match (avoids showing a bare separator)

Fixes #409

1. Open a CCloud-connected topic consumer
2. Open the settings panel — verify only `max.poll.records`, `fetch.max.bytes`, and `consumer.records.limit` are shown (no `request.timeout.ms`, `fetch.max.wait.ms`, or `max.partition.fetch.bytes`)
3. Set `max.poll.records` to 10 and consume — verify batches contain at most 10 records
4. Set `fetch.max.bytes` to a small value and verify it limits response sizes
5. Open a native Kafka consumer — verify all five Kafka config properties are still shown
6. Save and load presets — verify advanced settings persist correctly

## Pull request checklist

#### Tests

- [X] Added new
- [ ] Updated existing
- [ ] Deleted existing


src/io/confluent/intellijplugin/consumer/editor/KafkaConsumerSettings.kt
```diff
@@ -29,7 +46,7 @@ class KafkaConsumerSettings {
             ConsumerConfig.FETCH_MAX_WAIT_MS_CONFIG,
             ConsumerConfig.FETCH_MAX_BYTES_CONFIG,
             ConsumerConfig.MAX_PARTITION_FETCH_BYTES_CONFIG
-        ).forEach {
+        ).filter { it in supportedProperties }.forEach {
```
noeldevelops: Now that we have `ALL_PROPERTIES` defined up top, you can use it here to avoid double-declaring them all & having to update both in the future: `ALL_PROPERTIES.filter`...

src/io/confluent/intellijplugin/consumer/editor/KafkaConsumerPanel.kt
```diff
@@ -145,7 +146,14 @@ class KafkaConsumerPanel(
...
 
-    private val kafkaConsumerSettingsDelegate = lazy { KafkaConsumerSettings() }
+    private val kafkaConsumerSettingsDelegate = lazy {
+        val supportedProperties = if (kafkaManager is CCloudClusterDataManager) {
+            KafkaConsumerSettings.CCLOUD_PROPERTIES
+        } else {
+            KafkaConsumerSettings.ALL_PROPERTIES
+        }
+        KafkaConsumerSettings(supportedProperties)
+    }
```
noeldevelops, line 150: nit, non-blocking: this change makes the panel have to be connection-type-aware again, which isn't ideal imo
if we keep adding type-specific stuff maybe we just use a method like `getSupportedConsumerProperties()` on `BaseClusterDataManager` to keep the panel decoupled, kind of like we were with the `supportsAdvancedSettings`

src/io/confluent/intellijplugin/consumer/editor/KafkaConsumerSettings.kt
```diff
@@ -86,10 +103,12 @@ class KafkaConsumerSettings {
                 row(KafkaMessagesBundle.messageOrKey(it.key), it.value)
             }
 
-            separator()
+            if (propertiesFields.isNotEmpty()) {
+                separator()
```
noeldevelops, line 101: 👍 good attn to detail thanks

test/io/confluent/intellijplugin/consumer/editor/KafkaConsumerSettingsTest.kt
```diff
@@ -0,0 +1,109 @@
...
+        fun `should show empty properties for empty supported set`() {
+            val settings = KafkaConsumerSettings(emptySet())
+            val properties = settings.getProperties()
+            assertTrue(properties.isEmpty())
+        }
+    }
+
+    @Nested
+    @DisplayName("CCLOUD_PROPERTIES constant")
+    inner class CCloudPropertiesConstant {
```
noeldevelops: Let's just delete these tests? They seem pretty pointless since they just assert the constants contain what we defined in them.

test/io/confluent/intellijplugin/consumer/editor/KafkaConsumerSettingsTest.kt
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        @Test
+        fun `should have exactly 2 entries`() {
+            assertEquals(2, KafkaConsumerSettings.CCLOUD_PROPERTIES.size)
+        }
+    }
+
+    @Nested
+    @DisplayName("ALL_PROPERTIES constant")
```
noeldevelops: same as above... imo these aren't necessary and just a burden to reviewers/extra lines of code.

More useful tests would verify behavior like:
  - When CCLOUD_PROPERTIES is passed, getProperties() only returns values for the 2 supported keys
  - When ALL_PROPERTIES is passed, all 5 properties can be set and retrieved

noeldevelops (dismissed):
Advanced settings work great locally in click-testing! 

I'd love to see at least a couple unit tests for the core logic of resolving settings from `config.properties` and passing them to `ConsumeRecordsRequest`. 

Ok by me if that and other changes suggested here go into a follow-up PR for quicker iterating. 

For future design/ux improvements, maybe we don't need to hide these fields under Advanced Settings dialog, since there are so few fields here anyways, esp. compared to the direct connection type. 🤔

src/io/confluent/intellijplugin/consumer/editor/KafkaConsumerPanel.kt
bcscc: Thanks, just added the flag, a lot cleaner!

test/io/confluent/intellijplugin/consumer/editor/KafkaConsumerSettingsTest.kt
bcscc: you're right, updated them to actually set non-default values via `applyConfig()` and verify that `getProperties()` returns only the supported keys

test/io/confluent/intellijplugin/consumer/editor/KafkaConsumerSettingsTest.kt
bcscc: see above!

bcscc:
> Advanced settings work great locally in click-testing!
> 
> I'd love to see at least a couple unit tests for the core logic of resolving settings from `config.properties` and passing them to `ConsumeRecordsRequest`.
> 
> Ok by me if that and other changes suggested here go into a follow-up PR for quicker iterating.
> 
> For future design/ux improvements, maybe we don't need to hide these fields under Advanced Settings dialog, since there are so few fields here anyways, esp. compared to the direct connection type. 🤔

Thank for the feedback! I've made the updates in this PR and added a couple tests for the core logic. Will add further tests in a followup!

noeldevelops (approved):
tests look way better, thanks!

PR #439, opened 2026-03-05
Add partition filtering and optimized per-partition consume

### LINKED ISSUE #438: CCloud consumer: partition filtering and optimized per-partition consume
Add partition filtering support for CCloud consumer client and optimized consume to be per-partition when filter is active. 
- Add partition filter parsing, validation, and enforcement matching native consumer behavior
- Use per-partition offset resolution (`getPartitionOffset`) for filtered partitions instead of all-partition bulk fetches
- Use the single-partition GET endpoint for timestamp-based consume with a partition filter (POST `timestamp` applies to all partitions)
- Throw on `CONSUMER_GROUP` start type (not supported for CCloud)
- Move validation before `onStart()` so the UI doesn't grey out on error

bcscc:
### Summary of Changes

- Add partition filter support to `CCloudConsumerClient` matching native `KafkaConsumerClient` behavior
  - Parse and validate user-specified partitions against actual topic partitions via `describeTopicPartitions`
  - Intersect user input with real partitions; error with popup if no valid partitions remain
  - Validate synchronously before `onStart()` so the UI never greys out on validation failure
- Optimize initial consume for filtered partitions
  - Offset-based start types (BEGINNING, NOW, OFFSET, LATEST_MINUS_X) resolve offsets only for filtered partitions using `getPartitionOffset` in parallel, then POST with those offsets — avoids the N+1 API calls from `getTopicBeginningOffsets`/`getTopicEndOffsets` which hit ALL partitions
  - Timestamp-based start types (SPECIFIC_DATE, LAST_HOUR, TODAY, YESTERDAY) use the single-partition GET endpoint in parallel, since the POST `timestamp` parameter applies to all partitions
- Add single-partition consume endpoint infrastructure
  - `DataPlaneFetcher.consumePartitionRecords()` — single-partition GET with `timestamp`/`offset`/`max_poll_records` query params
  - `DataPlaneFetcher.getPartitionOffset()` — per-partition offset resolution
  - `SinglePartitionConsumeResponse` model and `CCLOUD_SINGLE_PARTITION_CONSUME_API_PATH` config
- Filter response records and track offsets only for filtered partitions in `pollLoop`
- Throw on `CONSUMER_GROUP` start type (not supported for CCloud) instead of silently falling back to NOW
- Extract `consumeRequest()` helper to eliminate duplication of `resolvedMaxPollRecords`/`resolvedFetchMaxBytes` across 7 call sites

Fixes #438

**Merge #434 first before this PR.**

1. Open a CCloud-connected topic consumer with a high-partition topic (50+ partitions)
2. Set partition filter to "0, 1" and consume from beginning — verify it starts quickly (should be ~3 API calls, not ~100+)
3. Verify records only come from partitions 0 and 1
4. Set partition filter to "999" (invalid) — verify a popup error appears and the UI does NOT grey out
5. Set start type to "Last Hour" with partition filter "0" — verify it uses single-partition GET and returns results
6. Set start type to "Consumer Group" — verify an error is shown
7. Consume without partition filter — verify behavior is unchanged
8. Test with `max.poll.records` set in advanced settings — verify it's respected for both filtered and unfiltered consume

**Performance improvement for filtered partition consume (100 total partitions):**

| Start type | Before (all-partition fetch) | After (filtered, e.g. 2 partitions) |
|---|---|---|
| BEGINNING / NOW | ~101 API calls (list + N offsets) | ~3 API calls (1 per partition offset + POST) |
| OFFSET / LATEST_MINUS_X | ~101 API calls | ~3 API calls |
| Timestamp types | 1 POST to all N partitions | 2 GET calls (1 per partition) |

## Pull request checklist

#### Tests

- [X] Added new
- [ ] Updated existing
- [ ] Deleted existing


src/io/confluent/intellijplugin/ccloud/config/CloudConfig.kt
```diff
@@ -63,6 +63,10 @@ object CloudConfig {
             /** Consume records with guaranteed progress per partition. */
             const val CCLOUD_SIMPLE_CONSUME_API_PATH =
                 "/kafka/v3/clusters/%s/internal/topics/%s/partitions/-/records:consume_guarantee_progress?return_raw_base64_records=true"
+
+            /** Consume records from a single partition (GET with query params). */
+            const val CCLOUD_SINGLE_PARTITION_CONSUME_API_PATH =
+                "/kafka/v3/clusters/%s/internal/topics/%s/partitions/%d/records"
```
flippingbits, line 75: I believe we can specify the partitions that we want to consume from in the request payload sent against the `CCLOUD_SIMPLE_CONSUME_API_PATH`. Could you double-check if it's possible? If so, manipulating the request payload might be easier than calling a second endpoint, also from a maintenance perspective since these are _internal_ endpoints which tend to change more often than public ones.

flippingbits (commented):
Thanks for optimizing the message viewer experience! I have one question on the endpoint being used.

src/io/confluent/intellijplugin/ccloud/config/CloudConfig.kt
bcscc: Good question! The `POST` endpoint's `offsets` field does support partition restriction and we already use it entirely for offset-based start types, but the server treats `offsets` and `timestamp` as [mutually exclusive](https://github.com/confluentinc/ce-kafka/blob/198768e6cece06f72a61c326794f0ee7815e04fe/ce-kafka-rest/ce-kafka-rest-extensions/src/main/java/io/confluent/kafkarest/resources/v3/SimpleConsumeAction.java#L204-L207). So we can't restrict specific partitions and seek by timestamp in a single `POST` call.

This PR adds the `GET` endpoint so that it can be called once to bootstrap the initial offsets for the filtered partitions. From then on, every poll uses the `POST` endpoint with those specific offsets.

Without the `GET` endpoint, the fallback is a `POST` using only a timestamp. The server then polls every partition on that first request, and we filter the data client-side. This overhead only happens once, and for small topics the cost is negligible. However, for large topics (100 partitions filtered to 2), the user experience may be noticeably slower.

However, I’m happy to consolidate to one endpoint if you think maintenance simplicity outweighs this initial performance hit. Let me know!

flippingbits: Makes sense, thank you! Let's keep it as is.

bcscc:
Thanks for the review @flippingbits, I've added coverage on the partition filter related methods will followup in another PR with more coverage on the poll(), start() and stop().

test/io/confluent/intellijplugin/consumer/client/CCloudConsumerClientTest.kt
```diff
@@ -872,4 +884,712 @@ class CCloudConsumerClientTest {
...
+            val filter = setOf(1)
+
+            whenever(mockFetcher.getPartitionOffset("test-topic", 1, false)).thenReturn(500L)
+            whenever(mockFetcher.consumeRecords(eq("test-topic"), any())).thenReturn(emptyResponse)
+
+            client.fetchInitialRecords(config, mockFetcher, "test-topic", partitionFilter = filter)
+
+            verify(mockFetcher).getPartitionOffset("test-topic", 1, false)
+            verify(mockFetcher, never()).getTopicEndOffsets(any())
+            Unit
```
flippingbits, line 1110: Out of curiosity, why is this line needed here?

flippingbits (dismissed):
Thanks for adding tests. LGTM!

test/io/confluent/intellijplugin/consumer/client/CCloudConsumerClientTest.kt
bcscc: So the Unit ensures the method's inferred return type stays Unit, which keeps JUnit 5 recognizing it as a valid test. Without it, the test would silently disappear from reports

noeldevelops (approved):

PR #444, opened 2026-03-06
feat(producer): Add CCloud REST produce endpoint and models (1/4)

bcscc:
## Summary of Changes

- **Add CCloud REST produce API request/response models** (`ProducerModels.kt`)
  - `ProduceRecordRequest`, `ProduceRecordData`, `ProduceRecordHeader`, `ProduceRecordResponse`, `ProduceRecordResponseData`
  - Uses `@Serializable` (kotlinx.serialization) matching the existing `ConsumerModels.kt` pattern
- **Add `PRODUCE_RECORDS_URI`** to `CloudConfig.DataPlane.Kafka`
- **Add `produceRecord()`** to `DataPlaneFetcher` interface and implementation
  - POST request with JSON-serialized body via `CCloudRestClient.executeRequest()`

### Feature Branch Stack (#413)
- **(this)** #444 — PR 1: Produce endpoint + request/response models
  - #456 — PR 2: Refactor producer client + provider setup
    -  #459 — PR 3: CCloud producer basic functionality (primitives)
       - #460 — PR 4: Full schema serialization support

Fixes #443

## Pull request checklist

Please check if your PR fulfills the following (if applicable):

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing

PR #456, opened 2026-03-06
feat(producer): Extract ProducerClient interface and add provider factory (2/4)

bcscc:
## Summary of Changes

- Extracted `ProducerClient` interface from `KafkaProducerClient` to support multiple producer implementations (native Kafka vs CCloud REST)
  - Mirrors the existing `ConsumerClient` / `ConsumerClientProvider` pattern used on the consumer side
- Created `ProducerClientProvider` factory that dispatches by `BaseClusterDataManager` type
  - `KafkaDataManager` → `KafkaProducerClient` (native protocol)
  - `CCloudClusterDataManager` → `CCloudProducerClient` (REST API, skeleton with `TODO()`)
- Updated `KafkaProducerEditor` to accept `BaseClusterDataManager` instead of `KafkaDataManager`
  - Hides native-only UI controls (compression, acks, idempotence) for CCloud connections via `isNativeConnection` flag
- Refactored `KafkaProducerClient` to use callback-based `onStart`/`onStop` pattern and moved lifecycle management (`isRunning`, `dispose`) into the client itself

### Feature Branch Stack
  -  **(this)** #456 — PR 2: Refactor producer client + provider setup
    - #459 — PR 3: CCloud producer basic functionality (primitives)
      - #460 — PR 4: Full schema serialization support

## Pull request checklist

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing


bcscc:
The failed coverage condition is due to SQ also including non-covered existing files in the code base. New code coverage is at ~83%

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -0,0 +1,50 @@
...
+        onUpdate: (Long, List<KafkaRecord>) -> Unit
+    ) {
+        TODO()
+    }
+
+    override fun stop() {
+        running.set(false)
+    }
+
+    override fun dispose() {
```
Cerchie, line 47: here, we should probably rename `stop()` to `dispose()`, since all `dispose()` does is call the stop func.

src/io/confluent/intellijplugin/producer/client/KafkaProducerClient.kt
```diff
@@ -28,15 +27,19 @@ import java.util.*
...
+class KafkaProducerClient(
+    val dataManager: KafkaDataManager,
+    val onStart: () -> Unit,
+    val onStop: () -> Unit
+) : ProducerClient {
+    val client = dataManager.client
     val connectionData = client.connectionData
 
-    val isRunning = AtomicBoolean(false)
+    internal val isRunning = AtomicBoolean(false)
```
Cerchie: nit: having two `isRunning` where one is a val and the other is a function bothers me, maybe rename the atomic boolean to `running` . However, this is pre-existing, won't block on it.

Cerchie (dismissed):
couple nits, overall though the code LGTM

src/io/confluent/intellijplugin/producer/client/KafkaProducerClient.kt
bcscc: Good call, renamed `isRunning` to `running`, thanks!

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
bcscc: You're right, for the producer, `dispose()` will always just call `stop()`. However, I think `stop()` may have to remain a separate function for the manual UI trigger, while `dispose()` satisfies the `Disposable` interface I factored out for UI reuse between native and CCloud clients.

Before, the editor used `Disposer.register(this) { it.isRunning.set(false) }` inline. Extending the interface with `Disposable` allows both client types to use the cleaner `Disposer.register(this, producerClient)` pattern. If you think there's is a better pattern, please let me know!!

Cerchie (approved):

PR #459, opened 2026-03-06
feat(producer): CCloud producer with primitive type support (3/4)

### LINKED ISSUE #414: CCloud Producer Client Basic Production
No CCloudProducerClient exists. KafkaCreateProducerAction sets isEnabled to false for CCloud. Create `CCloudProducerClient` that sends records via REST API. Support STRING key/value for initial scope. Return `KafkaRecord` with metadata (offset, partition, timestamp, duration).

bcscc:
## Summary of Changes

- Implement `CCloudProducerClient.produce()` with support for primitive types (STRING, JSON, LONG, INTEGER, DOUBLE, FLOAT, BASE64, NULL)
  - Builds `ProduceRecordRequest` with appropriate type encoding: STRING/JSON pass through as-is, numeric/binary types serialize to bytes and send as base64 BINARY
  - Handles manual/auto flow modes via `ProducerFlowParams`, CSV data frames, random data generation, and template processing
- Enable the "Produce" button for CCloud cluster connections in `KafkaCreateProducerAction.update()`
  - Previously only enabled for native Kafka (`isEnabled = isKafkaManager`), now also enabled for `CCloudClusterDataManager`
- Route CCloud connections through `CCLOUD_MANAGER_KEY` in `KafkaEditorProvider` and `KafkaCreateProducerAction.openProducer()`
  - `openProducer()` now accepts `BaseClusterDataManager` and sets the appropriate user data key based on manager type
- Schema serialization (SCHEMA_REGISTRY, AVRO_CUSTOM, PROTOBUF_CUSTOM) is deferred to PR 4

Fixes https://github.com/confluentinc/intellij/issues/414

### Feature Branch Stack (bcscc/ccloud-primitive-producer)
  - #456 — PR 2: Refactor producer client + provider setup
    - **(this)** #459 — PR 3: CCloud producer basic functionality (primitives)
      - #460 — PR 4: Full schema serialization support

1. Sign into a CCloud OAuth connection in the plugin
2. Select a CCloud environment and cluster 
3. Verify the "Produce" button is now **enabled** (was previously disabled for CCloud)
4. Click "Produce" to open the producer editor
5. Verify compression, acks, and idempotence controls are **hidden** for CCloud
6. Select a topic without a Schema, set key type to STRING, value type to STRING
7. Enter key and value text, click produce
8. Verify the record appears in the response table with correct topic, partition, offset
9. Test with other primitive types: LONG (e.g. `12345`), INTEGER, DOUBLE, FLOAT
10. Test AUTO mode with flow params (records per request, interval)
11. Verify native Kafka producer still works unchanged

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing


Cerchie:
2 edge cases that can be ticketed for later:

1. When I enter the value, `1.797693134862315E7308` in the Double key and val field for `Double.MAX_VALUE`, I get Infinity:

Screenshot 2026-03-10 at 8 48 05 AM

And while `1.7976931348623157 × 10³⁰⁸` is an astronomically large number, it's not infinity. 

2. This also happens when I enter `1.797693134862315E7308` in the Float key and val field. Since Floats hold fewer points of precision than doubles, I should get an error here like "Float can only hold ~7 decimal digits of precision" or similar

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -37,14 +65,258 @@ class CCloudProducerClient(
...
+                    partition = response.partitionId ?: -1,
+                    offset = response.offset ?: -1,
+                    duration = duration,
+                    timestamp = response.timestamp?.let { Instant.parse(it).toEpochMilli() } ?: System.currentTimeMillis(),
+                    keySize = response.key?.size ?: 0,
+                    valueSize = response.value?.size ?: 0,
+                    headers = processedHeaders,
+                    keyFormat = recordKey.schemaFormat,
+                    valueFormat = recordValue.schemaFormat,
+                    errror = null
```
Cerchie: `errror = null` typo?

bcscc: Thank you for the catch, fixed!

test/io/confluent/intellijplugin/producer/client/CCloudProducerClientTest.kt
```diff
@@ -1,35 +1,277 @@
...
+        fun `should build request with partition`() {
+            val key = createFieldConfig(KafkaFieldType.STRING, "k", isKey = true)
+            val value = createFieldConfig(KafkaFieldType.STRING, "v")
+            val request = client.buildProduceRequest(key, value,emptyList(), "test-topic", 5)
+
+            assertEquals(5, request.partitionId)
+        }
+
+        @Test
+        fun `should build request without partition when negative`() {
```
Cerchie, line 167: I think we should throw (or bubble up) [an error in this case](https://github.com/apache/kafka/blob/ccfc37d3743649128bcd606ad1d954ac76d12fcf/clients/src/main/java/org/apache/kafka/clients/producer/ProducerRecord.java#L77) in the underlying business logic for this test, based on what the record will do with it

Cerchie: Image

Cerchie (changes requested):
I ended up with 
```
Unhandled exception in Dispatchers.IO, interactive mode: false

java.lang.NoClassDefFoundError: com/intellij/charts/dataframe/DataFrameCSVAdapter
	at io.confluent.intellijplugin.util.csv.KafkaCsvUtils.readDataFrame(KafkaCsvUtils.kt:11)
	at io.confluent.intellijplugin.producer.client.CCloudProducerClient.produceLoop(CCloudProducerClient.kt:106)
	at 
```
When I tried producing with a CSV. 

I had to provide the charts class in META-INF for it to start working:
```
<dependencies>                                                                                                                                                                                     
 23 +        <module name="intellij.charts"/>                                                                                                                                                               
 24 +    </dependencies> 
```

Let's fix that, plus, let's not allow negative partition input

bcscc:
> I ended up with
> 
> ```
> Unhandled exception in Dispatchers.IO, interactive mode: false
> 
> java.lang.NoClassDefFoundError: com/intellij/charts/dataframe/DataFrameCSVAdapter
> 	at io.confluent.intellijplugin.util.csv.KafkaCsvUtils.readDataFrame(KafkaCsvUtils.kt:11)
> 	at io.confluent.intellijplugin.producer.client.CCloudProducerClient.produceLoop(CCloudProducerClient.kt:106)
> 	at 
> ```
> 
> When I tried producing with a CSV.

Hey @Cerchie, thanks for your review and comments! Do you mind if I address the CSV error in a followup PR because testing it against the native producer also causes the same error so i'm suspecting it's an pre-existing bug again. 

As for the negative partitions, the current native producer [overrides negative partition to null](https://github.com/confluentinc/intellij/blob/main/src/io/confluent/intellijplugin/producer/client/KafkaProducerClient.kt#L153-L160) which filters default to all partitions so I've made the CCloud producer to follow the same pattern. However I did add a new validation function to surface invalid inputted partitions. If you'd think we should change the current negative behavior, I can follow up in another PR.

Cerchie:
> Hey Lucia Cerchie (@Cerchie), thanks for your review and comments! Do you mind if I address the CSV error in a followup PR because testing it against the native producer also causes the same error so i'm suspecting it's an pre-existing bug again.
> 
> As for the negative partitions, the current native producer [overrides negative partition to null](https://github.com/confluentinc/intellij/blob/main/src/io/confluent/intellijplugin/producer/client/KafkaProducerClient.kt#L153-L160) which filters default to all partitions so I've made the CCloud producer to follow the same pattern. However I did add a new validation function to surface invalid inputted partitions. If you'd think we should change the current negative behavior, I can follow up in another PR.

>Do you mind if I address the CSV error in a followup PR because testing it against the native producer also causes the same error so i'm suspecting it's an pre-existing bug again.

Bummer! A follow-up would be fine, but be sure to `import com.intellij.charts.dataframe.DataFrame` at the top import section and write `csvDataFrame: DataFrame?` in L223 for this PR, instead of the inline type annotation. Doing that might even resolve the first layer of this issue on this branch ;) I think we'll need an update for Claude.md regarding style on that one as well, for a separate PR. 

I think we should think about the best user experience for the negative partition value. Let's bring it up at the next pairing session.

Cerchie (dismissed):

Cerchie (approved):

PR #460, opened 2026-03-06
feat(producer): Add schema serialization and retry with backoff (4/4)

### LINKED ISSUE #415: CCloud Producer Augmented Type Serialization
After support for STRING type, need remaining primitive types. Support INTEGER, LONG, DOUBLE, FLOAT, JSON, BASE64, NONE types. Map each to the correct REST API payload format. Reuse `KafkaFieldType` type selection and `ConsumerProducerFieldConfig.getValueObj()` for conversion.

bcscc:
## Summary of Changes

- Add client-side schema serialization for CCloud producer (SCHEMA_REGISTRY, AVRO_CUSTOM, PROTOBUF_CUSTOM)
  - SCHEMA_REGISTRY: fetches schema ID from SR, serializes payload with Avro/Protobuf/JSON, prepends Confluent V0 wire format prefix (magic byte + 4-byte schema ID), sends as base64 BINARY
  - AVRO_CUSTOM / PROTOBUF_CUSTOM: serializes payload without wire format prefix (no SR subject)
  - Mirrors the deserialization logic in `CCloudConsumerClient` (`getSchemaIdFromRawBytes`, `deserializeAvro`, `deserializeProtobuf`)
- Add retry with exponential backoff for transient produce errors
  - Retries on 429 (rate limit) and 5xx (server errors), up to 5 attempts
  - Exponential backoff: 1s → 2s → 4s → 8s → 16s, capped at 30s
  - Fails immediately on non-retryable 4xx client errors (400, 401, 403, 422, etc.)
- Cache Schema Registry schema IDs per subject for the lifetime of a produce session
   - `resolveSchemaCache()` pre-fetches schema IDs once at the start of `produceLoop()`, before entering the batch loop
   - The cache is passed through `produceBatch` → `buildProduceRequest` → `buildRecordData` → `buildSchemaRegistryData`
   - Falls back to a network call if a subject is not found in the cache
   - Cache is scoped to a single `start()` → `stop()` session — switching topics requires a new session, which gets a fresh cache
- Restore `suspend` and `fetcher` parameter on `buildRecordData` / `buildProduceRequest` (needed for SR lookups)

### Feature Branch Stack (bcscc/ccloud-primitive-producer)
  - #456 — PR 2: Refactor producer client + provider setup
    - #459 — PR 3: CCloud producer basic functionality (primitives)
      - **(this)** #460 — PR 4: Full schema serialization support

Fixes #415

 1. Sign into a CCloud OAuth connection with Schema Registry configured
 2. Open the producer editor for a CCloud cluster
 3. Select a topic with an Avro schema and verify the SCHEMA_REGISTRY is auto detected and filled.
 4. Enter valid record to produce matching the schema, click Produce
 5. Verify the record is produced successfully and appears in the consumer with correct deserialization
 6. Repeat with a Protobuf subject and a JSON Schema subject
 7. Test AVRO_CUSTOM: select AVRO_CUSTOM type, paste an inline Avro schema, enter matching data
 8. Verify it produces successfully
 9. Test retry: if possible, trigger a 429 rate limit (rapid-fire AUTO mode) and verify the producer recovers instead of failing
 10. Test AUTO mode with SR schema: run AUTO mode with multiple records and verify no rate-limiting issues from repeated SR lookups
 11. Verify primitive types (STRING, LONG, etc.) still work unchanged

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing


noeldevelops:
Not sure which part of the stack this belongs in, but I noticed that if I select a topic and click Producer, it doesn't open with that topic already selected, just the first in the list. (CCloud only, the original connection type does use the selected topic). 

IIRC you fixed a similar thing with Consumer before.

bcscc:
> Not sure which part of the stack this belongs in, but I noticed that if I select a topic and click Producer, it doesn't open with that topic already selected, just the first in the list. (CCloud only, the original connection type does use the selected topic).
> 
> IIRC you fixed a similar thing with Consumer before.

This should be fixed now!

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -308,6 +316,157 @@ class CCloudProducerClient(
...
+                response.errorCode
+            )
+
+            if (!isRetryableStatus(response.errorCode)) {
+                throw exception
+            }
+
+            lastException = exception
+            thisLogger().debug("Retryable produce error (attempt ${attempt + 1}/$MAX_RETRIES): ${response.errorCode}")
+
```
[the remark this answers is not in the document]
noeldevelops: Trying to understand this comment... I _think_ what it is saying is that the retry logic depends on `response` being returned, but the underlying call inside `produceRecord` is going to throw if it's a error response, not pass it along as `response`, so the retry logic won't get invoked. Does that sound right?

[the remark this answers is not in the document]
bcscc: Yup, `executeRequestInternal()` throws `CCloudApiException` for any non-2xx HTTP status before a response is ever returned, so 429/5xx errors would bypass the retry loop. Just fixed in the latest push by wrapping the call in a try/catch for `CCloudApiException` and applies the same retryability check to the `statusCode`. Added tests covering both HTTP-level and application-level error retry paths as well.

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -308,6 +316,157 @@ class CCloudProducerClient(
...
+        fetcher: DataPlaneFetcher,
+        field: ConsumerProducerFieldConfig
+    ): ProduceRecordData {
+        val schemaInfo = fetcher.getLatestVersionInfo(field.schemaName)
+        val schemaId = schemaInfo.id
+
+        val payloadBytes = when (field.schemaFormat) {
+            KafkaRegistryFormat.AVRO -> {
+                val avroSchema = field.parsedSchema as AvroSchema
+                val record = AvroSchemaUtils.toObject(field.valueText, avroSchema)
```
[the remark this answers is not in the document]
bcscc: Added schema caching.

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -308,6 +343,174 @@ class CCloudProducerClient(
...
+    }
+
+    /**
+     * Prepend the Confluent V0 wire format prefix to payload bytes.*
+     * Mirror of [CCloudConsumerClient.getSchemaIdFromRawBytes] which strips this prefix.
+     */
+    @VisibleForTesting
+    internal fun prependSchemaIdPrefix(schemaId: Int, payload: ByteArray): ByteArray {
+        val buffer = ByteBuffer.allocate(1 + 4 + payload.size)
+        buffer.put(0x00.toByte()) // magic byte
```
flippingbits: uber-nit: Could we replace this magic value with `io.confluent.kafka.serializers.schema.id.SchemaId.MAGIC_BYTE_V0`? It's not problematic at all in this case as the v0 magic byte will probably never change but try to use constants provided by dependencies in general to minimize maintenance work.

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -308,6 +343,174 @@ class CCloudProducerClient(
...
+                if (!isRetryableStatus(response.errorCode)) {
+                    throw exception
+                }
+
+                lastException = exception
+            } catch (e: CCloudApiException) {
+                // HTTP-level errors thrown by the REST client (e.g. 429, 5xx)
+                if (!isRetryableStatus(e.statusCode)) {
+                    throw e
+                }
```
flippingbits, line 501: Could we log the non-retryable exception/response, or is it getting logged somewhere else?

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -156,6 +177,21 @@ class CCloudProducerClient(
         }
     }
 
+    private suspend fun resolveSchemaCache(
+        fetcher: DataPlaneFetcher,
+        vararg fields: ConsumerProducerFieldConfig
+    ): Map<String, SchemaVersionResponse> {
+        val cache = mutableMapOf<String, SchemaVersionResponse>()
```
flippingbits: Is the returned/resolved cache used by multiple threads? If so, we should use a thread-safe structure, like [ConcurrentHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html).

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -308,6 +343,174 @@ class CCloudProducerClient(
...
+        field: ConsumerProducerFieldConfig,
+        schemaCache: Map<String, SchemaVersionResponse> = emptyMap()
+    ): ProduceRecordData {
+        val schemaInfo = schemaCache[field.schemaName]
+            ?: fetcher.getLatestVersionInfo(field.schemaName)
+        val schemaId = schemaInfo.id
+
+        val payloadBytes = when (field.schemaFormat) {
+            KafkaRegistryFormat.AVRO -> {
+                val avroSchema = field.parsedSchema as AvroSchema
```
flippingbits: I guess we can be sure that `field.parsedSchema` is always an `AvroSchema` if `field.schemaFormat` equals `KafkaRegistryFormat.AVRO`, thus no check of the type has to be performed before casting it? The same question applies to line 369.

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -308,6 +343,174 @@ class CCloudProducerClient(
...
+        val protobufSchema = field.parsedSchema as ProtobufSchema
+        val message = ProtobufSchemaUtils.toObject(field.valueText, protobufSchema) as DynamicMessage
+        return ProduceRecordData(
+            type = "BINARY",
+            data = Base64.getEncoder().encodeToString(message.toByteArray())
+        )
+    }
+
+    /**
+     * Prepend the Confluent V0 wire format prefix to payload bytes.*
```
flippingbits: nit: 
```suggestion
     * Prepend the Confluent V0 wire format prefix to payload bytes.
```

test/io/confluent/intellijplugin/producer/client/CCloudProducerClientTest.kt
```diff
@@ -347,4 +747,146 @@ class CCloudProducerClientTest {
...
+            partitionId = 0,
+            offset = 1
+        )
+
+        /** Set the private `running` AtomicBoolean to true so produceWithRetry doesn't short-circuit. */
+        private fun setRunning() {
+            val field = CCloudProducerClient::class.java.getDeclaredField("running")
+            field.isAccessible = true
+            (field.get(client) as java.util.concurrent.atomic.AtomicBoolean).set(true)
+        }
```
flippingbits: nit: Should this field be `@VisibleForTesting`?

flippingbits (commented):
Nice work! I have a few minor questions.

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
bcscc: Yes, parsedSchema is always set alongside schemaFormat in the field config but added an error to explicitly define the expected type.

noeldevelops (approved):

PR #463, opened 2026-03-10
fix: Prevent consumer dialog crash when clicking Presets after closing

### LINKED ISSUE #401: kafka Consumer dialog crash when clicking Presets toggle after closing
### Summary

[IncorrectOperationException](https://confluent.sentry.io/issues/7279637513/?project=4510160131588096&query=is%3Aunresolved&referrer=issue-stream)
Clicking the "Presets" expansion toggle immediately after closing the Kafka Consumer configuration dialog causes a crash.

### What Is Affected

Feature: Kafka Consumer configuration dialog
Component: Consumer Presets section (collapsible panel with saved configuration templates)
User Action: Clicking the expand/collapse arrow next to "Presets" heading

### Stack Trace
IncorrectOperationException: Parent `KafkaConsumerPanel` has already been disposed 
so child `ConsumerPresets` will never be disposed
```
  at io.confluent.intellijplugin.consumer.editor.KafkaConsumerPanel.presetsDelegate (KafkaConsumerPanel.kt:260)
  at io.confluent.intellijplugin.consumer.editor.KafkaConsumerPanel.getPresets (KafkaConsumerPanel.kt:268)
  at io.confluent.intellijplugin.core.ui.ExpansionPanel.setExpanded (ExpansionPanel.kt:120)
```

### User Journey (hard to reproduce)

1. Open Kafka tool window → New Consumer
2. Quickly click Cancel button
3. Immediately click the "▶ Presets" expansion arrow
4. Timing-dependent (more likely with slow UI response)

### Possible Root Cause

Timing issue: UI click event processed after dialog disposal, triggering initialization of component that needs to register with disposed parent.

noeldevelops:
## Summary
Fixes #401

- Added `!isDisplayable` guard to `ExpansionPanel`'s `expanded` setter so that click events processed after dialog disposal are silently ignored
- Prevents `IncorrectOperationException` when lazy `ConsumerPresets` component tries to register with an already-disposed `KafkaConsumerPanel`
- Protects all `ExpansionPanel` consumers, not just the Presets case

### Root Cause
The Presets `ExpansionPanel` starts collapsed with a lazy supplier (`{ presets.component }`). When the user closes the consumer dialog and immediately clicks the expand toggle, the click event is processed after `KafkaConsumerPanel` is disposed. The supplier triggers lazy initialization of `ConsumerPresets`, which calls `Disposer.register(this, presets)` on the disposed parent — throwing `IncorrectOperationException`.

### Why `isDisplayable`
`Component.isDisplayable()` returns `false` once the component is removed from a displayable containment hierarchy (i.e., after dialog disposal). The `init { update() }` block bypasses the setter, so initial setup is unaffected.

## Test plan
- [ ] Open Kafka tool window → New Consumer
- [ ] Quickly click Cancel, then immediately click the "▶ Presets" expansion arrow
- [ ] Verify no crash occurs


noeldevelops:
This bug was _really_ hard to reproduce, and I can't click fast enough to catch it. But the fix here is a solid guard that can only help imo.

bcscc (approved):
Having a hard time reproducing but LGTM from reading, thanks!

PR #477, opened 2026-03-11
chore: update to next version under development to `v253.25911.0`

noeldevelops:
## Summary of Changes

- Saw we left the `pre` suffix in plugin.xml before [here](https://github.com/confluentinc/intellij/pull/196/changes), so I left it


flippingbits (approved):
Thanks!

Issue #494, opened 2026-03-12
Custom protobuf schema not deserializing properly


bcscc:
consume from a message produced with/consumed by the same custom Protobuf schema causes an `Unsupported field type for byte deserialization: PROTOBUF_CUSTOM` error. expected to deserialize properly as the custom schema entered are the same.

<img width="2402" height="1088" alt="Image" src="https://github.com/user-attachments/assets/d0e5e985-2dd3-490e-a1fa-c70ca90ae9cb" />

PR #497, opened 2026-03-12
[v253.25910.x] fix deserialization of custom schemas 

Cerchie:
## Summary of Changes

Fixes #494 

Originally, no deserializer was being created in the `CCloudConsumerClient` for custom schemas. This caused the logic to fall through down to the logic for primitive types, in turn causing an error to be thrown even when the custom schema was the exact same one on the producer side. This fix threads it back through. 

On both CCloud Oauth'd connections and direct ccloud connections, test that:

1. you can produce to a topic with a custom Protobuf/Avro schema
2. you can consume from that same topic without the error reported in the ticket 

## Pull request checklist

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing


noeldevelops (approved):
thanks for the fix. Glad we caught this one! 
Confirmed this fixes the error locally. ✅

PR #492, opened 2026-03-12
[v253.25910.x] Change unsupported consumer/producer configs to disabled instead of hidden

bcscc:
## Summary of Changes

- Change unsupported CCloud OAuth consumer/producer config fields from **hidden** to **disabled** (grayed out)
  - Users can now see that options like compression, idempotence, acks, and advanced consumer properties exist but are unavailable for their connection type
- Producer: compression, idempotence, and acks rows are always rendered but disabled for CCloud connections
  - `SegmentedButton` (acks picker) requires explicit `component.isEnabled = false` after panel build since `Row.enabled()` doesn't propagate to its internal action buttons
- Consumer: all 5 advanced properties are always shown in the settings dialog, with unsupported ones disabled
  - `getProperties()` filters out unsupported properties to prevent sending them to the CCloud REST API
- Consumer: consumer group section changed from `.visibleIf()` to `.enabledIf()`, with `updateVisibility()` updated to respect `supportsConsumerGroups()`

Fixes https://github.com/confluentinc/intellij/issues/457

### Click-testing instructions

1. Open a CCloud OAuth connection in the plugin
2. Open the **Producer** editor for any topic
3. Expand the "Options" collapsible group
4. Verify **Compression**, **Idempotence**, and **Acks** (None/Leader/All) are visible but grayed out
5. Open the **Consumer** editor for any topic
6. Click "Advanced Settings" link
7. Verify all 5 consumer properties are shown, with `request.timeout.ms`, `fetch.max.wait.ms`, and `max.partition.fetch.bytes` grayed out
8. Expand "Other" settings and verify the **Consumer Group** field is visible but grayed out
9. Open a **native Kafka** connection and verify all fields remain fully functional (not disabled)

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing

Issue #457, opened 2026-03-06
Change unsupported consumer/producer configs to disabled instead of hidden


bcscc:
Change unsupported consumer advanced configs and producer configs to be disabled instead of hidden for CCloud OAuth Connections

PR #504, opened 2026-03-13
[v253.25910.x] Add 'not supported' tooltips to disabled CCloud OAuth fields

bcscc:
## Summary of Changes

- Add explanatory text to UI fields that are disabled for Confluent Cloud OAuth connections, so users understand *why* they are greyed out
  - Producer options (Compression, Idempotence, Acks): small text label rendered below each disabled field via `row { comment(...) }`
  - Consumer group selector: small text label below the disabled field
  - Consumer advanced settings dialog: hover tooltip on disabled `JTextField` inputs (replaces the Kafka config documentation tooltip for unsupported properties)

Follow-up to https://github.com/confluentinc/intellij/pull/492
Related to https://github.com/confluentinc/intellij/issues/457

1. Open a **Confluent Cloud OAuth** connection in the plugin
2. Open the **Producer** editor for any topic and expand the "Options" collapsible group
3. Verify Compression, Idempotence, and Acks fields are greyed out with *"Not supported by Confluent Cloud OAuth"* in small text beneath each
4. Open the **Consumer** editor, expand "Other" settings, and verify the Consumer Group field shows *"Not yet supported by Confluent Cloud OAuth"* beneath it
5. Click **Advanced Settings** link and hover over a greyed-out field (e.g. `request.timeout.ms`) — verify tooltip reads *"Not supported by Confluent Cloud OAuth"*
6. Open a **native Kafka** connection and verify no tooltip/label changes appear — all fields remain fully functional

After:

Screenshot 2026-03-16 at 5 02 30 PM
Screenshot 2026-03-16 at 5 02 19 PM
Screenshot 2026-03-16 at 5 02 11 PM

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


noeldevelops:
Let's wait on copy edits before merging this one

tikimims:
@noeldevelops @bcscc :) QQ: Is it possible to use hover tooltips on the disabled controls themselves? That way we could move away from having the duplicate inline help text. What do you think about the following approach:

For the disabled producer options, have a hover tooltip that is more generic and states the limitation instead of the internal workings (without mentioning OAuth or REST), like: 

- "Not configurable for this Confluent Cloud connection"

For the consumer group selector tooltip, how about: 

- "Consumer group selection isn’t available for this Confluent Cloud connection yet" (we could make this more concise if needed if there are contraints)

bcscc:
> Noel Cothren (@noeldevelops) Bernie Chen (@bcscc) :) QQ: Is it possible to use hover tooltips on the disabled controls themselves? That way we could move away from having the duplicate inline help text. What do you think about the following approach:
> 
> For the disabled producer options, have a hover tooltip that is more generic and states the limitation instead of the internal workings (without mentioning OAuth or REST), like:
> 
> * "Not configurable for this Confluent Cloud connection"
> 
> For the consumer group selector tooltip, how about:
> 
> * "Consumer group selection isn’t available for this Confluent Cloud connection yet" (we could make this more concise if needed if there are contraints)

@tikimims I agree, moving to hover tooltips sounds a lot cleaner! I can take on the fix.

Regarding the copy, I wanted to double-check something with you. Technically, these features can still be used for Confluent Cloud cluster if a user connects via the native protocol (direct cluster connection). They're only disabled in the new Confluent tab because the OAuth/REST API has these specific functional limits.

Do you think we should update to something like `"Not configurable using this connection, use a direct connection for full options."` copy to clarify this or do you think this is too much of a technicality and shouldn't cause a confusion to the user?

tikimims:
> Do you think we should update to something like `"Not configurable using this connection, use a direct connection for full options."` copy to clarify this or do you think this is too much of a technicality and shouldn't cause a confusion to the user?

@bcscc Thanks for clarifying. :) 

Could we make it clearer? For the disabled producer options, maybe like:

- "Not configurable using this connection. Connect directly to the Kafka cluster to change this setting."

And then for the consumer group selector tooltip, maybe like:

- "Consumer group selection isn’t available with this connection. Connect directly to the Kafka cluster to choose a group.”

bcscc:
> > Do you think we should update to something like `"Not configurable using this connection, use a direct connection for full options."` copy to clarify this or do you think this is too much of a technicality and shouldn't cause a confusion to the user?
> 
> Bernie Chen (@bcscc) Thanks for clarifying. :)
> 
> Could we make it clearer? For the disabled producer options, maybe like:
> 
> * "Not configurable using this connection. Connect directly to the Kafka cluster to change this setting."
> 
> And then for the consumer group selector tooltip, maybe like:
> 
> * "Consumer group selection isn’t available with this connection. Connect directly to the Kafka cluster to choose a group.”

@tikimims Sounds great, will make the updates, thank you for your help!

noeldevelops (approved):
🆒

PR #507, opened 2026-03-13
[v253.25910.x] fix: skip flush on producer stop to prevent ghost messages

### LINKED ISSUE #505: If producer is stopped too quickly, message doesn't show in producer but shows in consumer
https://github.com/user-attachments/assets/1ef586db-e552-4a67-b788-afe9af797457

bcscc:
## Summary of Changes

- Skip `producer.flush()` when producer is stopped and use `close(Duration.ZERO)` to abort buffered sends
  - When clicking "Produce" then "Stop" very fast, `flush()` in the `finally` block was force-sending buffered messages to the broker even though the producer had already been stopped, causing messages to appear in the consumer but not the producer panel

Fixes #505

1. Open a producer panel for any topic
2. Click "Produce" and then "Stop producing" as fast as possible
3. Verify the message does **not** appear in the consumer (it was never committed to the broker)
4. Produce normally (without immediately stopping) and verify messages still appear in both the producer and consumer panels

## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


noeldevelops (approved):
nice it was a simple fix 👍

PR #510, opened 2026-03-13
[v253.25910.x] Disable context menu for CCloud group in settings

bcscc:
## Summary of Changes

- Hide right-click context menu actions (Add, Remove, Duplicate) on the Confluent Cloud group node in the connections settings panel
  - Uses `fromPopup` flag from `createActions()` so toolbar buttons remain fully functional — only the popup menu items are hidden
- Fix bug where Message Broker group disappears when the last connection under it is deleted
  - `treeWalkUpRemove` now preserves top-level group nodes instead of removing them when they become empty

1. Open Settings → Confluent connections panel
2. Right-click on the Confluent Cloud group node
3. Verify the context menu does not* show Add, Remove, or Duplicate actions
4. Right-click on the Message Brokers group node
5. Verify the context menu does show actions as before
6. Verify the toolbar buttons (+, -, duplicate) still work normally for all nodes
7. Create a connection under an empty Message Brokers group, then delete it
8. Verify the Message Brokers group remains visible after deletion

## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


shouples (approved):
LGTM, curious if there's a way we could eventually get rid of this empty `Nothing here` action (that doesn't do anything, but also doesn't seem to hurt anything)
Image

bcscc:
> LGTM, curious if there's a way we could eventually get rid of this empty `Nothing here` action (that doesn't do anything, but also doesn't seem to hurt anything)

I did look into this but since the `Nothing here` option is the default behavior within the intellij IDE, it does end up being a more involved fix if we want to change it.

Issue #508, opened 2026-03-13
Pre-filling topics upon consumer/producer open is not consistent when topic list is stale


bcscc:
1) open a direct connection (not Oauth) to the clicktesting CC env 
2) create a new topic in CC 
3) before it's refreshed,  try to create a producer or consumer off a topic and see it get the new topic name instead of the correct one

PR #516, opened 2026-03-16
Add Claude preference: no @Suppress for type issues

Cerchie:
## Summary
- Adds a code style rule to CLAUDE.md: don't use `@Suppress` annotations to work around type issues — find a properly typed solution instead
- Triggered by a correction during #508 fix where an unchecked cast with `@Suppress("UNCHECKED_CAST")` was replaced with a properly typed approach

## Test plan
- [ ] Verify CLAUDE.md renders correctly


Cerchie:
closing, need to fix this local claude skill I have to check out a new branch

PR #533, opened 2026-03-17
[v253.25910.x] fix: Fast produce cancellation causing ghost records

### LINKED ISSUE #527: Fast produce cancellation causing ghost records
KafkaProducerClient: Fast produce cancellation with multiple records in a batch causes leaked produced records that don't show up in the producer message viewer. 

CCloudProducerClient: Fast produce cancellation with single (similar to #505) or multiple records batch cause same bug as above ^

Additional: super fast produce processing (can be recreated with local connection) causes multiple produce even while button state is still loading.

Fix flushing logic for producer clients for batching, and refactor cancel behavior to be consistent across both client types.

bcscc:
## Summary of Changes

- **Fix ghost records on fast stop** — records produced to the broker/CCloud but never shown in the producer UI
   - `KafkaProducerClient`: once `producer.send()` is called, always waits for the broker ack (bounded by existing 15s send timeout) instead of cancelling the future on stop — since `Future.cancel()` only cancels the ack wait, not the actual send
   - `CCloudProducerClient`: wraps REST produce call in `NonCancellable` so the HTTP call always completes, avoiding ambiguous record state on job cancellation
 - **Incremental UI updates** — both clients now call `onUpdate` per record instead of batching, so confirmed records appear immediately and are never lost on early stop
 - **Fix double-produce on fast click** — disables the produce button immediately on EDT before launching async work, re-enables on `onStart`/`onStop`/validation failure
 - **Unified stop behavior** — both clients now defer `onStop()` to the `finally` block of `start()`, showing a loading/stopping state while  in-flight work completes. `stop()` only sets the running flag — it does not trigger UI changes directly

Fixes #527

1. Open a producer editor for any topic (both native Kafka and CCloud connections)
2. Set mode to **Manual** with batch count > 1, click Produce, immediately click Stop
3. Verify all records that were confirmed by the broker appear in the output table — no ghost records
4. Set mode to **Auto** with a short interval, click Produce, immediately click Stop
5. Verify all confirmed records appear and the button shows a brief loading/stopping state before resetting
6. Double-click Produce rapidly — verify only one produce session starts
7. Verify normal produce (without fast stop) still works unchanged

#### Tests

- [x] Updated existing
- [x] Added new


src/io/confluent/intellijplugin/producer/client/KafkaProducerClient.kt
```diff
@@ -257,25 +252,42 @@ class KafkaProducerClient(
...
+                keyConfig = correctKey, valueConfig = correctValue,
+                metadata = metaInfo, duration = (end - start),
+                headers = formedHeaders
+            )
+        }
+
+        // Stopped — wait briefly for the in-flight send to complete to avoid ghost records
+        if (!isRunning()) {
+            return try {
+                val metaInfo = metadataFuture.get(2, TimeUnit.SECONDS)
```
Cerchie: I'm not sure how this would perform on different networks. @noeldevelops what do you think of this approach? 

If we keep it, can we extract `TimeUnit.SECONDS` to a named constant?

noeldevelops: @bcscc Can you clarify what happens here if the `send` doesn't complete after 2 sec? will it throw?

bcscc: If the send doesn't complete in 2 secs, a TimeoutException will be thrown which is then caught by the catch and we then cancel() and return null, so the record is treated as unconfirmed and not shown in the UI. However since the producer's sender thread is independent to this process, ghost messages could still happen if the broker does receive and commit the in-flight record. 

The alternative is to remove the 2 sec grace period and let the producer just always wait for an in-flight send to complete (only fall back to the 15s timeout limit) and have `Stop` only prevents starting new sends via the `isRunning()` check in the for-loop. @Cerchie @noeldevelops If y'all think that may be a better approach? Only tradeoff is the UI responsiveness of a fast `Stop` will be dependent on the time it takes to spin up a kafka producer client and send at least one message.

bcscc: I've made an update to go with the alternative, however please let me know if y'all have any different ideas! Thanks

src/io/confluent/intellijplugin/producer/client/KafkaProducerClient.kt
```diff
@@ -257,22 +251,19 @@ class KafkaProducerClient(
...
 
-        if (!isRunning())
+        if (!metadataFuture.isDone) {
+            metadataFuture.cancel(true)
             return null
-        val metaInfo = metadataFuture.get(2, TimeUnit.SECONDS)
-        val end = System.currentTimeMillis()
+        }
 
+        val metaInfo = metadataFuture.get()
```
noeldevelops: I think we could make this code a little simpler: we don't need the whole timeout loop here to wait on completion, esp since we don't check `isRunning` inside it anymore. 

`Future.get(timeout)` does the same thing but uses the future's internal wait/notify mechanism instead of needing to poll with `Thread.sleep(100)`

Maybe you could  replace all these lines with something like:
```
  val metadataFuture = producer.send(record as ProducerRecord<Any, Any>)                                                                                                                                         
                                                                                                                                                                                                                 
  val metaInfo = try {                                                                                                                                                                                           
      metadataFuture.get(15, TimeUnit.SECONDS)                                                                                                                                                                   
  } catch (e: TimeoutException) {                                                                                                                                                                                
      metadataFuture.cancel(true)                                                                                                                                                                                
      return null                                                                                                                                                                                                
  }     
```

src/io/confluent/intellijplugin/producer/client/CCloudProducerClient.kt
```diff
@@ -91,9 +90,9 @@ class CCloudProducerClient(
         }
         onStart()
 
-        producerScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
+        val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
```
noeldevelops: Pasting in Claude's comment about this below. I agree it could be slightly cleaner, but it's a `nit`

> val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
>   produceJob = scope.launch { ... }
> 
>   The scope is a local variable that's never cancelled. The old code cancelled it in stop(). The new code only cancels the produceJob. This works because the scope has only one child job, but it's sloppy.
>   Consider either:
>   - Storing the scope and cancelling it in stop() (cleaner lifecycle)
>   - Or just using produceJob = CoroutineScope(Dispatchers.IO).launch { ... } and letting the scope die with the job

src/io/confluent/intellijplugin/producer/editor/KafkaProducerEditor.kt
```diff
@@ -229,6 +229,8 @@ class KafkaProducerEditor(
     }
 
     private fun startProduce(): Boolean {
+        produceButton.isEnabled = false
```
noeldevelops, line 232: I see this line of code, but the button doesn't look/act disabled to me when testing? IDK if that's a bug in this code or just an annoying thing we need to fix in the button UI, or maybe just me?

noeldevelops (commented):
Testing locally this does fix the problem! 

I see it removes the batch processing on the producer - good move, that really was the root cause of ghost records. If we ever see a big lag in UI updates for large produce batches we can implement a fix on the UI side instead. 

I left comments for a couple minor changes to make it a little cleaner.

src/io/confluent/intellijplugin/producer/editor/KafkaProducerEditor.kt
bcscc: The disable prevents double-click immediately, it's re-enabled by onStart() and switches to the `Stop` button, so the gap may be too brief to see visually. For Kafka producer client this gap includes the broker connection setup (~1s), but for CCloud it may be too brief to notice

bcscc:
@noeldevelops Thanks for the review, I've made the changes!

src/io/confluent/intellijplugin/producer/editor/KafkaProducerEditor.kt
noeldevelops: That makes sense. I still think there's no visual difference? Or maybe it's the focus outline that confuses me. Either way no fix needed atm.

noeldevelops (approved):
TY!

PR #534, opened 2026-03-17
bypass "Trust and Open Project ___" dialog when using `runIde`

shouples:
Removes this before opening the `intellij` (or custom, if defined) workspace window:

image |


noeldevelops (approved):
🙏

Issue #531, opened 2026-03-17
Refresh CCloud connection resets the entire view


noeldevelops:
### Steps to Reproduce
1. Sign in to Confluent cloud
2. Expand your view to a specific cluster, then topic, the the Topic's key or value schema tab: 
<img width="1079" height="262" alt="Image" src="https://github.com/user-attachments/assets/852c03d0-3b2b-4892-8cf6-a1a03d6b67eb" />

3. Click "Refresh" in the top right

5. **Observe**: after a pause while loading (lack of a loading indicator could be a [separate issue](https://github.com/confluentinc/intellij/issues/514)), the view on the right is reset to the top-level Env:<img width="1399" height="263" alt="Image" src="https://github.com/user-attachments/assets/d5f44f71-8666-4132-9f5e-61a6fcba2d84" />
**Expect**: the view should stay on the tab I was looking at

flippingbits:
Fixed by https://github.com/confluentinc/intellij/pull/538

Issue #425, opened 2026-03-03
DataPlaneCache does not update cachedTopics on create/delete


bcscc:
`DataPlaneCache.createTopic()` and `DataPlaneCache.deleteTopic()` do not update `cachedTopics` after successful API calls. This causes stale reads from `getTopics()`, which makes the consumer panel's topic dropdown inconsistent when a new topic is created while the panel is open.

Improve test coverage of`DataPlaneCache`. Add `DataPlaneCacheTest` covering:
- `createTopic` adds topic to cache and `getTopics()` reflects it
- `deleteTopic` removes topic from cache and `getTopics()` reflects it
- `refreshTopics` replaces full cache
- `getTopics` returns empty list when cache is null (before any refresh)
- Cache state after `dispose()` is cleared

noeldevelops:
See also - Schema Cache create and delete operations. Could be done with this? cc @bcscc

PR #538, opened 2026-03-17
[v253.25910.x] Fix CCloud refresh resetting view to top level

eguicf:
## Summary of Changes

  - Fixed https://github.com/confluentinc/intellij/issues/531
  - Refresh button now only refreshes the currently visible panel (cluster/topic/schema) instead of refreshing
  everything
  - View stays on the current panel after refresh instead of resetting to environment/root view
  - Schema types/versions now show loading indicators ("...") during refresh
  - Cancels ongoing jobs immediately when refresh is clicked
  - **Improve refresh code logic** (tried to refresh only the current panel, but changed back to refresh button refresh all data (expected))

**Main Issue - View stays on current panel after refresh**                                   
  1. Open CCloud connection:
    - Add a CCloud connection
    - Sign in and wait for the tree to load
  2. Navigate to a nested resource:
    - Expand an environment 
    - Select a Kafka cluster/SR cluster OR
    - Expand a Kafka cluster, click on a specific topic OR
    - Expand Schema Registry and click on a specific schema OR
    -  click on a key/value schema tab under a topic detailed panel: steps stated [here](https://github.com/confluentinc/intellij/issues/531)
  3. Trigger refresh:
    - Click the refresh button in the toolbar
  4. Observe:
    - ❌ Before fix: View jumps back to the environment level (top level)
    - ✅ After fix: View stays on the selected view

**other issue noticed during the change - Refresh not work on Schema/Topic detail panel** 
1. go to ccloud schema detail tab or key/value schema tab. 
2. create/delete/update schema or topic data externally (on CCloud UI)
3. Click refresh button                                                                     
4. panel should display the updated schema/topic, the tree view should be updated as well.
5. the improvement of refresh logic is not breaking any else/bringing back old issues, like  https://github.com/confluentinc/intellij/issues/425

## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing


src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentMainController.kt
```diff
@@ -307,8 +308,7 @@ internal class ConfluentMainController(
 
         myTree.addTreeSelectionListener {
             val treePath = it.newLeadSelectionPath
-            if (treePath != null)
-                lastSelectedPath = treePath
+            lastSelectedPath = treePath
```
noeldevelops: Should we preserve the null check here, in case the treePath is nullified during the refresh somehow?

noeldevelops (commented):
In testing this locally I don't see the refresh taking me back to the environments/top-level _if_ I'm viewing a specific topic, but if I have a cluster selected and click Refresh, it still goes back to the 'root' view after all the topics load 🫥 (Only oauth connection tested)

https://github.com/user-attachments/assets/549e4c43-75bf-4c2c-92d5-6666f947e0ad

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentMainController.kt
eguicf: ~~yes, we should check for null, the update in [L95](https://github.com/confluentinc/intellij/pull/538/changes/BASE..f6c732a4b4f1a68753a286e7bf6df848358b2f86#diff-abca28a54a4a22c10326735660724510c02ff455e6d7b06737d83adef844db0fR95) `private var lastSelectedPath: TreePath? = null` directly enabled the simplification from the if-check~~ 

update: correct, nice catch, the previous one with if null check in the function is better, reverted.

src/io/confluent/intellijplugin/ccloud/cache/DataPlaneCache.kt
```diff
@@ -127,6 +127,10 @@ class DataPlaneCache @VisibleForTesting internal constructor(
 
     fun getSchemas(): List<SchemaData> = cachedSchemas ?: emptyList()
 
+    fun clearSchemaCache() {
```
eguicf, line 134: This is something I missed earlier. Added`clearSchemaCache` to force fresh API fetch during refresh. Without this, schemas were using cached data.

src/io/confluent/intellijplugin/consumer/client/CCloudConsumerClient.kt
```diff
@@ -748,7 +749,7 @@ class CCloudConsumerClient(
     private suspend fun fetchAndParseSchema(
         registryId: SchemaRegistryClusterId,
         key: SchemaCacheKey,
-        fetch: suspend () -> io.confluent.intellijplugin.ccloud.model.response.SchemaByIdResponse
+        fetch: suspend () -> SchemaByIdResponse
```
eguicf, line 752: unrelated to issue fix, but changed FQN to top-level import (code style consistency).

eguicf:
> In testing this locally I don't see the refresh taking me back to the environments/top-level _if_ I'm viewing a specific topic, but if I have a cluster selected and click Refresh, it still goes back to the 'root' view after all the topics load 🫥 (Only oauth connection tested)
> 
>  root.view.after.refresh.mov

Thank you for the thorough click-testing and finding out new issues @noeldevelops. Above issue is fixed.

src/io/confluent/intellijplugin/registry/common/KafkaSchemaInfo.kt
```diff
@@ -12,6 +12,7 @@ import java.util.*
 data class KafkaSchemaInfo(
     val name: String,
     @field:CustomRendering(RegistryFormatRenderer::class)
+    @field:LoadingRendering
```
eguicf, line 15: Added `@LoadingRendering` to type field so it shows "..." when null during refresh. Now both type and version show loading indicators.

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
```diff
@@ -41,6 +43,10 @@ internal class KafkaRegistryController(
     val registryType = dataManager.registryType
     private val model: ObjectDataModel<KafkaSchemaInfo> = dataManager.schemaRegistryModel!!
 
+    private val cardLayout = CardLayout()
```
eguicf, line 56: Not really related, but quickly fixed in this file: an [old fix](https://github.com/confluentinc/intellij/pull/540/changes#diff-abca28a54a4a22c10326735660724510c02ff455e6d7b06737d83adef844db0fR598) for CCloud empty schema table (showing custom empty panel with "Create Schema" link) was incorrectly triggering for non-empty schemas. Now it only worked for empty schema table.

existing issue for non-empty schema panel
Screenshot 2026-03-19 at 12 59 07 AM



now only apply for empty schema panel
Screenshot 2026-03-19 at 12 44 18 AM

src/io/confluent/intellijplugin/rfs/ConfluentDriver.kt
```diff
@@ -80,6 +80,12 @@ class ConfluentDriver(
         clusters.forEach { cluster ->
             registerClusterTopicListener(cluster.id, cluster)
         }
+
+        // Also register schema registry listener
```
eguicf, line 84: Added schema registry listener registration alongside topic listeners. Ensures both topics and schemas auto update in tree when data changes.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentMainController.kt
```diff
@@ -98,6 +98,13 @@ internal class ConfluentMainController(
     private lateinit var component: JComponent
     private lateinit var lastSelectedPath: TreePath
 
+    /** Get currently selected path for refresh operations */
```
eguicf: Removed lines of complex old refresh logic that was causing panel resets. 
Added simple `getCurrentSelectedPath()` helper for refresh button to know what to refresh.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentSchemaDetailController.kt
```diff
@@ -146,14 +146,6 @@ internal class ConfluentSchemaDetailController(
         updateVersion2Info()
     }
```
eguicf, line 153: Removed unused `refresh()` method.
Detail panels now update automatically when data refreshes (no separate refresh call needed).

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentTabController.kt
```diff
@@ -116,11 +116,6 @@ class ConfluentTabController(
 
     internal fun getMainController(): ConfluentMainController? = resourceController
```
eguicf, line 118: Removed unused `refreshDetailPanel()` method.
We now only refresh data models; detail panels update automatically.

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
```diff
@@ -111,21 +115,34 @@ class KafkaMonitoringToolWindowController(project: Project) : MonitoringToolWind
...
-
-            driver?.let {
-                it.dataManager.updater.stopAll()
-                it.dataManager.cancelAllEnrichmentJobs()
-
-                tabController.getMainController()?.refreshControlPlane()
-
-                it.safeExecutor.coroutineScope.launch {
-                    it.dataManager.updater.reloadAll(checkConnection = false)
+            val tabController = getConfluentCloudTabController() ?: return
```
eguicf: Changed refresh button to only refresh currently visible panel.

noeldevelops: I am curious about this -- Does this change make it only refresh the data on a single tab/panel? 

I can see the logic for doing it that way, but it seems like the refresh button should refresh _all the data_ for that connection (or at least clear the caches so the data will get fetched when the user tries to view it again). 

What do you think? Would refreshing all the data be a really heavy operation that we want to avoid?

eguicf: Appreciate raising the concern! 
~~The current approach refreshes only the visible panel to minimize API calls, because refreshing all the data can be heavy and much slower especially for large numbers of topic and schema enrichments. 
But this is something we should fix, and align with Kafka (direct connection) refresh button. I will make an issue for it, testing "refresh all data behavior", or maybe try a middle option.~~
update: now refresh button refresh all data.

eguicf:
@noeldevelops Thank you for testing again. 
The topic cache wasn't being cleared on refresh, so loadTopics() always returned stale cached data. Added clearTopicCache() and call it before refreshing (mirroring the existing schema refresh pattern). 
Now If we create a new topic in the ccloud UI then refresh the topic panel, the new topic will display.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentMainController.kt
```diff
@@ -503,7 +458,6 @@ internal class ConfluentMainController(
 
             val cluster = dataManager.getCachedKafkaClusters(envId)?.find { it.id == clusterId }
             if (cluster == null) {
-                myTree.clearSelection()
```
noeldevelops, line 506: I see several of these `myTree.clearSelection()` were removed, and wonder if it could cause stale selected items - if during update the selected item is removed (e.g., cluster deleted), the tree node would remain selected but the detail panel will show an empty placeholder?

I'm not sure if I'm reading the flow right and that could happen or not?

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
```diff
@@ -86,21 +90,36 @@ class KafkaMonitoringToolWindowController(project: Project) : MonitoringToolWind
...
+                    // Schema Registry selected → refresh only schemas
+                    val cluster = driver.dataManager.getCachedKafkaClusters(selectedEnvId)?.firstOrNull() ?: return
+                    val clusterDataManager = driver.dataManager.getOrCreateClusterDataManager(cluster) as CCloudClusterDataManager
+                    // Clear cache to force real refresh (not cached data)
+                    clusterDataManager.getDataPlaneCache().clearSchemaCache()
+                    clusterDataManager.updater.invokeRefreshModel(clusterDataManager.schemaRegistryModel!!)
                 }
             }
 
-            tabController?.refreshDetailPanel()
```
noeldevelops, line 103: was this removed intentionally? It might be what's preventing schema details from refreshing when I'm viewing a topic's schema tab

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
```diff
@@ -76,47 +82,69 @@ internal class KafkaRegistryController(
...
-        if (dataTable.parent is JViewport) {
-            dataTable.emptyText.attachTo(dataTable, dataTable)
+        if (!hasSchemas && !hasFilters) {
+            updateEmptyPanel()
+            cardLayout.show(wrapperPanel, "EMPTY")
+        } else {
+            cardLayout.show(wrapperPanel, "TABLE")
+            (decoratedTableComponent.layout as? BorderLayout)
+                ?.getLayoutComponent(BorderLayout.NORTH)?.isVisible = hasSchemas || hasFilters
+            dataTable.tableHeader?.isVisible = hasSchemas || hasFilters
```
noeldevelops: nit: this could be simplified... 
in the else block ln115-116 you're setting things visible based on `hasSchemas || hasFilters`, but the if condition `!hasSchemas && !hasFilters` is the logical opposite, meaning those checks will always return `true`

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
```diff
@@ -86,21 +90,36 @@ class KafkaMonitoringToolWindowController(project: Project) : MonitoringToolWind
...
+            val tabController = getConfluentCloudTabController() ?: return
+            val driver = tabController.getDriver() ?: return
+            val selectedEnvId = driver.selectedEnvironmentId ?: return
+            val selectedPath = tabController.getMainController()?.getCurrentSelectedPath()
+
+            // Cancel ongoing jobs immediately
+            driver.dataManager.updater.stopAll()
+            driver.dataManager.cancelAllEnrichmentJobs()
+
+            // Only refresh the currently visible data model
```
noeldevelops: I see cluster and schema path checks, but what about when Refresh is clicked when a Topic or Env is selected? Seems like it will just silently fail/do nothing, and needs an `else` to refresh all? 

Maybe that is what we'll solve with this issue though: https://github.com/confluentinc/intellij/issues/557

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
```diff
@@ -86,21 +90,36 @@ class KafkaMonitoringToolWindowController(project: Project) : MonitoringToolWind
...
+                    clusterDataManager.getDataPlaneCache().clearTopicCache()
+                    clusterDataManager.updater.invokeRefreshModel(clusterDataManager.topicModel)
+                }
+                selectedPath?.isSchemaRegistry(driver) == true -> {
+                    // Schema Registry selected → refresh only schemas
+                    val cluster = driver.dataManager.getCachedKafkaClusters(selectedEnvId)?.firstOrNull() ?: return
+                    val clusterDataManager = driver.dataManager.getOrCreateClusterDataManager(cluster) as CCloudClusterDataManager
+                    // Clear cache to force real refresh (not cached data)
+                    clusterDataManager.getDataPlaneCache().clearSchemaCache()
+                    clusterDataManager.updater.invokeRefreshModel(clusterDataManager.schemaRegistryModel!!)
```
noeldevelops: Since the type definition says `clusterDataManager.schemaRegistryModel` could be `null`, it'd be safer to  use `?.let { ... }` instead of `!!` here:

`clusterDataManager.schemaRegistryModel?.let { clusterDataManager.updater.invokeRefreshModel(it) }`

Otherwise if it ends up `null` at runtime we'll get an exception thrown

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
```diff
@@ -76,47 +82,69 @@ internal class KafkaRegistryController(
                 getSelectedItem()?.name?.let { KafkaDriver.schemasPath.child(it, false) }
         }
 
-        updateEmptyText()
+        wrapperPanel.add(emptyPanel, "EMPTY")
+        wrapperPanel.add(super.getComponent(), "TABLE")
+
+        model.addListener(object : DataModelListener {
```
noeldevelops: not sure: should this listener be registered as a Disposable so it doesn't leak?

noeldevelops (commented):
thanks for working on this Ellie it is quite complex !

I noticed one issue when click-testing this time: If I'm viewing the schema detail tab for a topic it doesn't update the schema data on refresh (won't show a new schema I added in Cloud, and if I evolve the schema it doesn't show the latest one).

Might be related to one of my comments below about `tabController?.refreshDetailPanel()` removal? 

Left some other comments/questions.

eguicf:
Revisited the previous improvement of refresh only the current panel, that's not a good approach and brought more issue. so I reverted back to refresh all data in the connection, 
`it.dataManager.getAllClusterDataManagers().forEach { clusterDataManager ->`
also fixed [this](https://github.com/confluentinc/intellij/issues/557).

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
eguicf: yes, updated.

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
eguicf: nice suggestion, simplified and removed redundant boolean check

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
eguicf: reverted the refresh logic to refresh all, that makes more sense.

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
eguicf: that's correct. With improved logic, we are now using `?.let` and null-safe

src/io/confluent/intellijplugin/ccloud/cache/DataPlaneCache.kt
```diff
@@ -98,6 +98,10 @@ class DataPlaneCache @VisibleForTesting internal constructor(
 
     fun getTopics(): List<TopicData> = cachedTopics ?: emptyList()
 
+    fun clearTopicCache() {
```
eguicf, line 101: added `clearTopicCache()` / `clearSchemaCache()` / `clearAllVersionCaches()` - Force fresh API fetch instead of returning stale cached data

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
```diff
@@ -76,47 +92,66 @@ internal class KafkaRegistryController(
...
-            }
+            cardLayout.show(wrapperPanel, "TABLE")
+            (decoratedTableComponent.layout as? BorderLayout)
+                ?.getLayoutComponent(BorderLayout.NORTH)?.isVisible = true
+            dataTable.tableHeader?.isVisible = true
+            if (!hasSchemas) updateEmptyText()
+        }
+    }
+
+    private fun updateEmptyPanel() {
```
bcscc, line 128: nit: used only in `updateUIForEmptyState`, pretty short function could be collapsed inline.

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
```diff
@@ -76,47 +92,66 @@ internal class KafkaRegistryController(
...
+            KafkaMessagesBundle.message("schemas.empty.text.create.link"),
+            SimpleTextAttributes.LINK_ATTRIBUTES
+        ) {
+            KafkaRegistryAddSchemaDialog(project, dataManager).show()
         }
     }
 
-    override fun emptyTextProvider() = CustomEmptyTextProvider {
-        updateEmptyText()
+    private fun updateEmptyText() {
```
bcscc, line 139: nit: same here, only used in `updateUIForEmptyState`, could be collapsed inline.

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
```diff
@@ -7,6 +7,7 @@ import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Disposer
+import com.intellij.openapi.diagnostic.thisLogger
```
bcscc:
```suggestion
```
nit: unused imports

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
```diff
@@ -86,21 +87,34 @@ class KafkaMonitoringToolWindowController(project: Project) : MonitoringToolWind
...
+                    }
+                    clusterDataManager.schemaRegistryModel?.let { model ->
+                        clusterDataManager.updater.invokeRefreshModel(model)
+                    }
+
+                    val versionModels = clusterDataManager.schemaVersionModels.getModelsForRefresh()
+                    versionModels.forEach { model ->
+                        clusterDataManager.updater.invokeRefreshModel(model)
+                    }
+                }
```
bcscc: It seems clearing the caches here (`clearTopicCache()/clearSchemaCache()`) is causing the refresh to reset the topic/schema list scroll position and selection due to a brief null state. Could we instead move the cache refresh into the background coroutine so that it updates directly from the old data?

bcscc (commented):
Thanks for the fix, I am able to see that the refresh no longer reset to the top-level Env however found the loading experience a bit buggy with scroll position being reset in the topic list as well as some slow loading.

https://github.com/user-attachments/assets/32b08f28-b8d1-456e-9fd2-d2846d563d11

We're getting close but if we can smooth out the experience more that would be great!

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
eguicf: good suggestion, but should be ok to keep them separate, to be clearer.

src/io/confluent/intellijplugin/registry/confluent/controller/KafkaRegistryController.kt
eguicf: It’s okay to keep `updateEmptyText` separate, since it’s also defined in `TopicsController.kt`

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
eguicf: removed, and a few more unused imports

src/io/confluent/intellijplugin/toolwindow/KafkaMonitoringToolWindowController.kt
eguicf: great catch, and suggestion! now scroll position should be preserved at refresh.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentMainController.kt
```diff
@@ -347,6 +347,7 @@ internal class ConfluentMainController(
         driver.selectedEnvironmentId = envId
         myTree.clearSelection()
 
+        dataManager.updater.stopAll()
```
eguicf, line 350: Added `updater.stopAll()` to cancel ongoing refresh when switching environments.

noeldevelops (approved):
This is testing really well. Previous issues with stale data are gone. 

I like the refresh all pattern, even if it's a little slower. We can discuss as a team how best to optimize the loading time for a follow up task.

PR #549, opened 2026-03-19
[v253.25910.x] fix: CCloud producer double-encoding JSON values

### LINKED ISSUE #542: JSON doesn't deserialize correctly in Consumer, Export
### Steps to reproduce:
1. Connect to cluster (*not sure if this is connection-type dependent)
2. Set up a new topic, do not set up a schema for it
3. Use the plugin's Producer, and set the Value schema type to `JSON`
4. Set the Flow to randomly generate key/values
5. Produce a message (or more than one)
6. Open a Consumer for the topic, set the Value schema type to `JSON`
7. Consume `From the Beginning` (or `Today` or `Last Hour`) to see the message(s) you just produced.
8. **Observed**: Consumer table shows all values with ugly newline and escape characters, e.g. `{\n \"id\": 0 }`. **Expected**: the values match what was displayed in Producer table, clean JSON. 
9. Export as JSON, it's even worse. (Could be related to [this Export bug](https://github.com/confluentinc/intellij/issues/235)?)

Image

bcscc:
## Summary of Changes

- Fix CCloud producer sending double-encoded JSON values by using REST API type `STRING` instead of `JSON` for schemaless JSON
  - `ProduceRecordData.data` is `String?`, so kotlinx.serialization string-escapes JSON content when `type = "JSON"`, producing values like `{\n \"id\": 0 }` instead of clean JSON
  - Mirrors the native `KafkaProducerClient` which uses `StringSerializer` for JSON type

Fixes #542

1. Connect to a CCloud cluster
2. Create a topic without a schema
3. Open Producer, set Value schema type to `JSON`, set Flow to generate random key/values
4. Produce one or more messages
5. Open Consumer for the same topic, set Value schema type to `JSON`
6. Consume from the beginning
7. Verify values display as clean JSON (no `\"` escape characters or `\n` newlines)

#### Tests

- [ ] Added new
- [x] Updated existing
- [ ] Deleted existing


noeldevelops (approved):
fixes the issue when testing locally!

PR #574, opened 2026-03-24
[do not merge] table rendering benchmarks

noeldevelops:
## Overview

- Contains tooling for comparing JBTable vs JCEF table rendering performance within the IDE. Originally built to validate that the JCEF-based consumer table performs acceptably at various data sizes.
- Built with Claude's help but not without a hard-won battle to control and understand the outcome. 

Important details about the benchmarks:
- Synthetic data generates "realistic" fake Kafka records. It uses a fixed random seed for reproducible results across runs
- JBTable benchmarks measure `paintImmediately` instead of waiting on the usual thread queue for a more direct comparison of rendering times. By default JBTable will only render the rows that are visible with a small buffer. 
- The HTML table implements virtual scrolling ("lazy loading") since that's closer to the way JBTable works ^; only the rows visible in the viewport (plus a buffer of 30 rows) are rendered as DOM nodes. 
     - Virtual scrolling means the larger tests never actually try to render all, for example, 100k rows, since they won't be visible on screen, **so the HTML benchmark is actually measuring rendering performance for visible rows _plus_ the time it takes to pass a large number of records** from Kotlin to JS. 

### Benchmark Results
- You can run this locally and join the fun! View `JCEF_PERFORMANCE_BENCHMARK_README.md` for details
Screenshot 2026-03-24 at 2 17 18 PM

#### Key findings

- `JBTable` time is constant ~1-9ms even with larger row counts bc the data path is zero-serialization; table uses Java object references in the same heap.
- `JCEF` has a ~20ms fixed floor (IPC round-trip + double-requestAnimationFrame), plus a cost that scales linearly with data size (JSON serialization of all data, even though only some records get DOM nodes).
- `JCEF`'s bottleneck is the data transfer path, not DOM rendering. With virtual scrolling, DOM work is comparable, but every record must cross the JVM-Chromium process boundary as serialized data. **If we implement this for our Consumer, we could use pagination to control the number of records we need to pass over the wire or render in the browser client.**
- JBCefJSQuery#addHandler callbacks can freeze JCEF rendering if not dispatched to background tasks, even though they appear to be off-EDT in the debugger. Confusing! 

I asked Claude how Webviews in VSCode are different. Here's what it gave me. 🤔  

> The VS Code extension's webview approach works well for different reasons than IPC speed — both JCEF and Electron use the same underlying Chromium IPC (Mojo). The key differences are: (1) JCEF adds a JNI boundary (Java <-> native CEF) and forces data embedding in executeJavaScript strings (requiring base64 encode/decode), while Electron's postMessage uses structured clone directly; (2) VS Code has no Swing modality constraints that cause deadlocks; (3) Electron's architecture treats webviews as first-class citizens rather than bolting a browser onto a Java UI framework. The fundamental serialization bottleneck (every record must be serialized to cross the process boundary) exists in both.


noeldevelops:
no longer needed

PR #600, opened 2026-04-13
fix: auto-size details panel key/value editors based on content height

noeldevelops:
Cherry pick of #595 into patch release branch


eguicf (approved):
LGTM, let's merge after CI approved.

PR #603, opened 2026-04-15
chore: prep changelog and version for patch release v253.25910.1

noeldevelops:
## Summary of Changes

- Update version and changelog for today's release


resources/META-INF/plugin.xml
```diff
@@ -2,7 +2,7 @@
     <id>com.intellij.bigdatatools.kafka</id>
     <name>Confluent</name>
     <vendor>Confluent</vendor>
-    <version>253.25910.0-pre</version>
+    <version>253.25910.1-pre</version>
```
noeldevelops, line 5: Is this right? the `pre` throws me off not sure how that works

.versions/next.txt
```diff
@@ -1 +1 @@
-253.25910.0
\ No newline at end of file
+253.25910.1
```
noeldevelops, line 1: Clarifying: even though it says "next", it should be the version we're releasing?

eguicf: It seems yes from [previous](https://github.com/confluentinc/intellij/commit/172400675406c03a9daa841066c3aaf6d576ee3c) `v253.25909.4` pattern. I think we just want to make sure `.versions/next.txt` in main branch set to the actual next version

eguicf (approved):

PR #605, opened 2026-04-17
Resolve schema references for display and update in CCloud

### LINKED ISSUE #602: [CCloud] Schema viewer does not resolve or display schemas with references
### Summary
In CCloud, the schema viewer fails to display it if a schema contains references to another schema. 
Direct and local Schema Registry connections handle this correctly.

### Steps to reproduce
1. Connect to a CCloud environment in the plugin
2. Create a schema reference to another schema using CLI

### Actual behavior
Schema fails to display:

Image

or produce:

Image

### Scope
CCloud connection: affected
Direct SR connection: working
Local SR connection: working
Image

eguicf:
## Summary of Changes
This PR closes https://github.com/confluentinc/intellij/issues/602.
In CCloud, opening a schema that references another schema (e.g. User → Address) threw an NPE and left a blank panel.
**Fix:** Introduce `preResolvedRefs` — a `refName → schemaText` map fetched up front via the CCloud SR REST API — and pass it to the providers.

## Changes

- CCloudClusterDataManager.kt — `parseSchemaForDisplay` and `updateSchema` pre-fetch all refs (including transitive) via the new `fetchResolvedReferences` helper (recurses with a cycle guard).
- KafkaRegistryUtil.kt — new `parseSchema` overload that accepts the pre-resolved map.
- Bdt{Avro,Json,Protobuf}SchemaProvider.kt — optional `preResolvedRefs` constructor param; when null, falls back to resolveReferences(schema), so Direct/Local paths are unaffected.
- ConfluentSchemaDetailController.kt:
  - Parse on a background thread before flipping the UI.
  - New `hasError` row — load failures now show a message instead of a blank panel, and re-clicking retries.
  - Removed a duplicate `updateVersion1Info()` from the version-combo callback that caused a "nothing to show" flash on ref schemas.

In ccloud, in `clicktest-env`, use `ellie-xx-user`schema to test display and produce/consume.

Screenshot 2026-04-20 at 10 00 52 AM
Screenshot 2026-04-20 at 10 00 44 AM
Screenshot 2026-04-20 at 10 00 37 AM
Screenshot 2026-04-20 at 10 01 56 AM

## Pull request checklist

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing


src/io/confluent/intellijplugin/registry/KafkaRegistryUtil.kt
```diff
@@ -75,6 +75,27 @@ object KafkaRegistryUtil {
...
+     * content has already been fetched from the CCloud SR REST API.
+     *
+     * @param resolvedReferences map of reference name → schema string (may be empty for schemas with no references)
+     */
+    fun parseSchema(
+        schemaType: KafkaRegistryFormat,
+        newText: @NlsSafe String,
+        references: List<SchemaReference>,
+        resolvedReferences: Map<String, String>
+    ): Result<ParsedSchema> {
```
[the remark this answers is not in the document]
eguicf: 👍  tests added.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentSchemaDetailController.kt
```diff
@@ -157,19 +161,22 @@ internal class ConfluentSchemaDetailController(
...
                     isLoading.set(false)
                     hasContent.set(true)
+                    hasError.set(false)
                 }
                 return@onSuccess
             }
 
             version1Schema = versionInfo
             val prettySchema = KafkaRegistryUtil.getPrettySchema(schemaType = versionInfo.type.name, schema = versionInfo.schema)
+            val parsedSchema = dataManager.parseSchemaForDisplay(versionInfo).getOrNull()
```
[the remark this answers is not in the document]
eguicf: updated.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentSchemaDetailController.kt
```diff
@@ -157,63 +155,61 @@ internal class ConfluentSchemaDetailController(
 
         isLoading.set(true)
         hasContent.set(false)
+        hasError.set(false)
 
         dataManager.getSchemaVersionInfo(schemaName, version).onSuccess { versionInfo ->
             if (version1Schema == versionInfo) {
                 invokeLater {
                     if (Disposer.isDisposed(this@ConfluentSchemaDetailController)) return@invokeLater
+                    if (schemaName != this@ConfluentSchemaDetailController.schemaName) return@invokeLater
```
eguicf: Added this line to skip if the user switched to a different schema while this fetch was in flight.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentSchemaDetailController.kt
```diff
@@ -114,6 +108,10 @@ internal class ConfluentSchemaDetailController(
             label(message("confluent.cloud.details.schema.loading")).align(Align.CENTER)
         }.resizableRow().visibleIf(isLoading)
 
+        row {
+            label(message("confluent.cloud.details.schema.load.failed")).align(Align.CENTER)
```
eguicf, line 112: Distinguish "load failed" from "empty schema" when the panel goes blank.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentSchemaDetailController.kt
eguicf: fixed this in [a follow up PR](https://github.com/confluentinc/intellij/pull/606) as well as renaming functions like `updateVersion1Info`(and vars `version1/version2`) which seems to be too vauge.

> Refactor this method to reduce its Cognitive Complexity from 32 to the 15 allowed.
> Check failure: Refactor this method to reduce its Cognitive Complexity from 32 to the 15 allowed.
> [S3776] Cognitive Complexity of functions should not be too high

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentSchemaDetailController.kt
```diff
@@ -79,10 +75,12 @@ internal class ConfluentSchemaDetailController(
...
+            // Do not call updateVersion1Info() directly — updateComboBox's first addItem fires
+            // onChanged which already triggers it. Calling again races the fast-path against
+            // the slow ref-resolving parse and flashes the empty panel.
             version1.component.item = versions.first()
-            updateVersion1Info()
         } else {
-            isLoading.set(false)
+            isLoading.set(true)
             hasContent.set(false)
         }
```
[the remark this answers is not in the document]
bcscc: nit: Could address in a followup, the no version found / no state case should probably have a separate UI than loading so that users don't get confused if the content is still loading or hit an error state.

src/io/confluent/intellijplugin/data/CCloudClusterDataManager.kt
```diff
@@ -528,16 +530,48 @@ class CCloudClusterDataManager(
...
+    private suspend fun fetchResolvedReferences(
+        references: List<SchemaReference>,
+        visited: MutableSet<String> = mutableSetOf()
+    ): Map<String, String> {
+        if (references.isEmpty()) return emptyMap()
+        val fetcher = dataPlaneCache.getFetcher() ?: return emptyMap()
+        val result = mutableMapOf<String, String>()
+
+        for (ref in references) {
+            if (!visited.add(ref.name)) continue
```
bcscc: We should use something like`"${ref.subject}:${ref.version}"` to prevent against the case that the name of a referenced schema is shared between two different schemas

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentSchemaDetailController.kt
```diff
@@ -54,9 +52,6 @@ import javax.swing.JCheckBox
 import javax.swing.JComponent
 import javax.swing.JPanel
 
-/**
- * CCloud schema detail view with tree/raw toggle, version selector, and compare mode.
```
bcscc, line 58: nit: keep doc string

src/io/confluent/intellijplugin/data/CCloudClusterDataManager.kt
```diff
@@ -528,16 +530,48 @@ class CCloudClusterDataManager(
...
+            versionInfo.references,
+            resolvedRefs
         )
     }
 
+    /**
+     * For each reference pointer, fetch the actual schema text via the CCloud SR REST API.
+     * Recurses into transitive refs and returns a flat `refName → schemaText` map.
+     */
+    private suspend fun fetchResolvedReferences(
```
bcscc: Can we add some unit tests to cover this new function?

bcscc (dismissed):
Thanks Ellie! Click testing well, just some a few commits

eguicf:
> Thanks Ellie! Click testing well, just some a few commits

@bcscc Thanks for the comments! ~~will address them in a follow up PR.~~ Edit: addressed them in this PR.

src/io/confluent/intellijplugin/toolwindow/controllers/ConfluentSchemaDetailController.kt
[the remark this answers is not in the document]
eguicf: Updated. Couldn't just set `isLoading = true` for empty because the callback fires twice (once before the fetch starts, again after it finishes), and both calls looked identical with the old signature. Made the callback parameter nullable `(List<T>?)` so we can tell "still loading" (null) apart from "loaded, no versions" (empty list).

src/io/confluent/intellijplugin/data/CCloudClusterDataManager.kt
eguicf: Good point, updated to use `subject:version`. But this only partially fixes the issue though: the resolved-references map is still keyed by name, so when two refs share a name only one wins. Verified in `clicktesting-env` with `ellie-shared-name-report`. And the same (existing) bug already exists in the direct connection. 
Seems not a simple fix, so we can handle the full fix in a separate PR.
Screenshot 2026-04-29 at 5 02 15 PM
Screenshot 2026-04-29 at 5 02 05 PM
Screenshot 2026-04-29 at 5 01 57 PM

src/io/confluent/intellijplugin/data/CCloudClusterDataManager.kt
eguicf: yes, those has been added. 
fyi, most of the [30 "new" lines in](https://sonarqube.confluent.io/component_measures?id=intellij&pullRequest=605&metric=new_coverage&view=list) `KafkaSchemaController.kt` are renames (version1Controller → selectedVersionController, updateVersion1Info → updateSelectedVersion, etc.), not new logic. Sonar counts touched lines as "new"; but nothing changed there.

bcscc (dismissed):
Thanks Ellie!

bcscc (approved):

PR #614, opened 2026-04-28
test: add unit tests for core util classes

flippingbits:
## Summary
- Add unit tests for `TimeUtils`, `SizeUtils`, `StringUtils`, `ExceptionUtils`, and `BdtUrlUtils` — pure-logic helpers in `core/util` that previously had no tests.
- ~60 new tests across 5 files, covering branch-heavy paths (interval formatting, byte unit boundaries, URL default ports, exception cause walking).
- First step in raising overall test coverage from the current ~12% baseline; targets utilities first since they're easy to isolate.


test/io/confluent/intellijplugin/core/util/BdtUrlUtilsTest.kt
```diff
@@ -0,0 +1,123 @@
...
+import com.intellij.testFramework.junit5.TestApplication
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNotNull
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+@TestApplication
+@DisplayName("BdtUrlUtils")
```
Cerchie, line 10: It will be especially good to have these if we eventually introduce some sort of intellij protocol similar to the built-in one we have in vscode for opening projects within the IDE

test/io/confluent/intellijplugin/core/util/BdtUrlUtilsTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+import com.intellij.testFramework.junit5.TestApplication
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNotNull
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+@TestApplication
```
Cerchie:
```suggestion
```

For here and the `StringUtilsTest`, Claude suggested we didn't need to boot up the application as the logic was separate and this can improve performance. Seems so: I removed `@TestApplication` from both files and it still worked.

Cerchie (changes requested):
Image

🔥 looking much better already in terms of coverage! 

I think we can remove `@TestApplication` from BdtUrlsUtilsTest and StringUtilsTest though

test/io/confluent/intellijplugin/core/util/BdtUrlUtilsTest.kt
flippingbits: Fair point! I removed the annotation from both files in 113b42b

flippingbits:
@Cerchie Thanks for your feedback! I've addressed it. Could you take another look when you have time?

Cerchie (approved):
LGMT upon CI passing!

PR #617, opened 2026-04-30
perf: CircularBuffer base class

noeldevelops:
## Summary
This is a new base `CircularBuffer` class for Consumer. No callers yet. 

### Why a new CircularBuffer class?
- it's more performant at eviction than `ArrayList`
- we need a stable index for items even after wrapping unlike `ArrayDeque` (for upcoming work on Search/Filter/Histogram)
- Circular buffer + bitset for filters is a proven performant data structure combo in our other message viewers (VSCode, Confluent Cloud UI)

## Implementation Details
Adds a fixed-capacity ring buffer with stable slot indices. `append` returns the slot the value was written into and any evicted predecessor; `removeHead` enables soft-cap eviction without buffer wrap; iteration yields entries oldest→newest.

The iterator snapshots `head` and `size` at construction so concurrent mutation (`append` / `removeHead` / `clear` mid-iteration) does not silently return stale or wrong-slot elements.

No callers in this PR — pure additive new file.

## Stack

This PR is the bottom of a 5-PR stack:
- **PR A: this** — `CircularBuffer<T>`
- PR B: `ConsumerRecordIndex` over slot indices
- PR C: `ListTableModel` backed by `CircularBuffer`
- PR D: wire `ConsumerRecordIndex` into `KafkaRecordsOutput`
- PR E: `SearchBarController` BitSet refactor


src/io/confluent/intellijplugin/consumer/data/CircularBuffer.kt
```diff
@@ -0,0 +1,103 @@
...
+ * primitive. The owning component should document where mutations happen and how readers see
+ * a consistent view.
+ */
+class CircularBuffer<T : Any>(val capacity: Int) : Iterable<T> {
+
+    init {
+        require(capacity > 0) { "capacity must be positive, got $capacity" }
+    }
+
+    private val slots: Array<Any?> = arrayOfNulls(capacity)
```
Cerchie, line 25: Note: this Any type leads to 4 suppressions, at L44, 53, 66 and 93. 

There's also an `Any` at L15

Cerchie (changes requested):
If there's a possibility of replacing these two `Any`s I'd do it; otherwise looks good

noeldevelops:
@Cerchie - thanks for flagging those suppressions !
For line 21 the alternative option that could be more specific - MutableList - will give us a performance hit down the road when using this class to render the table. So I kept Array<Any?> storage, but made a private accessor so we can make the unchecked cast in one call site, and everything else goes through there. 

The class declaration line 15 is safe - this `T : Any` means "T must be a non-nullable type." It's not a value type but a way to bound the type param

Cerchie (approved):

PR #619, opened 2026-04-30
perf: ListTableModel uses CircularBuffer 

noeldevelops:
## Summary
Finally, a user-facing change! 
`ArrayDeque` was the previous solution to making this table more performant, but it does not have the stable slot indexes that we want for upcoming filtering and Histogram. This replaces it with the `CircularBuffer` class we just added. User-facing changes are minimal - one mention in CHANGELOG for the default consumer limit getting capped in this PR (going from unbounded/unlimited to default 10k rows, unless a different limit is specified in config)

###  Click-testing options:
1. Smoke — Open a consumer connection, consume any topic, verify rows display.
2. Wrap behavior — Easiest way: temporarily edit BUFFER_CAPACITY = 20 in KafkaRecordsOutput.kt, consume a topic with >20 records. After wrap: click each column header to sort. Should reorder cleanly, no exception, no missing/ghost rows.
3. Soft cap — Set "Consumer records limit" = 5, consume a high-throughput topic. Verify only the latest 5 rows visible at any time; older rows evict as new arrive.
4. Clear button — With data loaded, click Clear; table empties, no exception.
5. Producer panel — Open a producer panel, send messages, verify they appear in its records table
6. Long-session — Let a high-throughput consumer run long enough to wrap naturally (10K+ records). Verify the UI stays responsive, no NPE in the IDE log, stats counter keeps updating.

### AI-generated Summary
PR 3 of 5. Replaces the `ArrayDeque` backing store inside `ListTableModel` with the new `CircularBuffer<T>`. JBTable's `AbstractTableModel` contract requires contiguous row indices `0..rowCount-1`; the buffer's slot indices are not contiguous after wrap. The translation `slot = (head + row) % capacity` lives entirely inside the model and is exposed via `fun slotForRow(row: Int): Int` so a future filter layer can key BitSets by slot — slots are stable across repaints, row indices are not.

#### Behavior preserved

Carries forward everything PR #431 won:
- Batched `flushPendingAdds` — one `fireTableRowsInserted` per flush, no per-element events
- O(1) random access (`ArrayDeque` was already a circular buffer internally; this just makes slot indexing explicit)
- O(1) eviction
- `replaceAll`, `elements()`, `clear()` semantics intact

#### What's new

- New optional ctor parameter `onSlotChange: (slot, next: T?) -> Unit` — defaulted to no-op so producer panel and other callers compile unchanged. PR D uses this to drive `ConsumerRecordIndex`. Two call shapes:
  - **append (fresh or wrap-overwrite)** → `(slot, next)` with `next != null`
  - **pure eviction (soft cap, no replacement)** → `(slot, null)`

  Consumers that need to react to the prior occupant track it themselves keyed by slot — see `ConsumerRecordIndex`'s inverse mapping for the canonical pattern.

- Wrap vs. soft-cap fire different table events: soft-cap eviction emits `fireTableRowsDeleted` + `fireTableRowsInserted` against a consistent `rowCount`; wrap (buffer at hard cap) emits a single `fireTableDataChanged` since every existing row index shifts.

- Soft-cap eviction (when `maxElementsCount < buffer.capacity`, e.g. `setMaxRows(10)` on a 1000-slot buffer) is implemented via `CircularBuffer.removeHead()`.
- New `addClearListener(action: () -> Unit)` — fired from both `clear()` and `replaceAll()` so external indices can stay in sync.

Constructor signature changed from `(data: MutableList<T>, ...)` to `(capacity: Int, ...)`. Three call sites updated, all one-line changes:
- `KafkaRecordsOutput.kt`
- `ListTableModelTest.kt`
- `ConsumerEditorUtilsTest.kt`

### Stack

- PR A: CircularBuffer (#617, merged)
- PR B: ConsumerRecordIndex (#618)
- **PR C: this**
- PR D: wire index into KafkaRecordsOutput (#620)
- PR E: SearchBarController BitSet refactor (#621)
  
### Test plan

- [x] Extended `ListTableModelTest` with wrap, slot-translation, eviction-callback, and TableModel-event cases
- [x] All existing batching, capacity, and replaceAll tests stay green
- [x] Producer panel + presets tests pass (the defaulted `onSlotChange` keeps them compiling unchanged)
- [x] `./gradlew test` — full suite green


resources/messages/KafkaBundle.properties
```diff
@@ -80,7 +80,7 @@ toggle.details=Details
 consumer.table.awaiting=Awaiting data
 # Kafka consumer advanced settings and properties fields
 consumer.records.limit=Consumer records limit
-consumer.records.limit.descr=Old records will be cleaned up on client when the records count reaches the limit.<br>Negative, '0' or empty value means no limit.
+consumer.records.limit.descr=Old records will be cleaned up on client when the records count reaches the limit.<br>Maximum 10,000 records are retained regardless of this setting. '0', empty, or negative values use the 10,000-record default.
```
noeldevelops: Open Q: should we boost the max records retained to 50k ?

`ConsumerRecordIndex` allocates `LongArray(capacity) + IntArray(capacity) + BitSet(capacity)` eagerly, so memory cost scales linearly with whatever number we choose, but users who are used to consuming a large number or messages or leaving the consumer running for a long time may not appreciate the 10k cap.

Cerchie: Should we test it to see if 50k negatively impacts performance? 

If 10k is the band, that might be all right, if I were just scrolling through, but considering that we can search messages, I'd think an increase would be appropriate

Cerchie:
thanks for the clicktesting instructions 

on step 6, hm, I'm seeing some jumpiness/freeziness at high throughputs... thoughts? 

https://github.com/user-attachments/assets/1d3a9c5d-fa99-4502-8573-40369e91df6c

resources/messages/KafkaBundle.properties
flippingbits: +1 on considering/testing it. 10k feels quite low.

noeldevelops: thanks y'all. I'll test it out and address @Cerchie 's comment below to make sure we're not seeing glitches after we start wrapping at large numbers

noeldevelops:
@flippingbits @Cerchie thanks for the review suggestions & jumpiness report

I learned that once the `CircularBuffer` filled, every flush forced the table to rebuild (and to drop selection/scroll) on _every_ EDT tick. Updated to evict with a sliding window (delete-top + insert-bottom) instead of rebuilding. 
 
Also found that sorting on the string columns (esp. value) became unusable with large message payloads. My solution here is to disable sorting those columns for now - this is not something high load data tables often support (including our vscode message viewer). We can reenable when Paused if there's a demand, but will want to optimize the string comparison function (swap `Collator` for `String.compareTo/naturalOrder`). 
 
Also, I did some benchmarking for the default 10k limit and bumped it to 50k bc the difference was not bad on responsiveness (heap scales linearly with payloads but this is expected). Branch `ncothren/perf-list-table-model-benchmark`. Write up [here](https://github.com/confluentinc/intellij/compare/ncothren/perf-list-table-model-buffer...ncothren/perf-list-table-model-benchmark?short_path=650b6e7#diff-650b6e781c3a05090b07f2d0d7231c541f3ed09e6131c6df7d7542b04225534c)

Cerchie (dismissed):
smooth message flow with high throughput topic 👏🏻 

also verified that sorting was disabled for string type cols

noeldevelops:
FYI merged the next one down into this. No changes, already approved

Cerchie (approved):

PR #621, opened 2026-04-30
perf: SearchBarController routes through a slot-keyed BitSet

noeldevelops:
## Summary
Global search feature was added recently with decent performance outcomes. In this PR we optimize it using the BitSet pattern established in [Confluent Extension for VS Code](https://github.com/confluentinc/vscode/)'s message viewer, and we avoid a costly RowFilter on every repaint. Contains a new `FreeTextSlotIndex` class that owns the search query text and BitSet. 

### Click-testing
1. Run plugin with UI or  `./gradlew runIde`
2. Search for a term in the columns or global search while consuming past the buffer wrap point of 50k messages
  - [ ] Expect that matches stay correct and newly-arrived matching records appear
  - [ ] Expect that changing the query and/or clearing the search updates the table responsively 

### AI-generated Summary
PR 5 of 5. The prototype's free-text `RowFilter` ran `String.contains` on every visible row on every repaint. This PR replaces it with an **incrementally-maintained, slot-keyed `BitSet`** — the same approach the Confluent VS Code extension's `Stream` uses (full scan once per term, then O(1) maintenance per inserted message).

#### `FreeTextSlotIndex`

A new index (sibling to `ConsumerRecordIndex`) owns the active term + a slot-keyed `BitSet`:

- **Term change** (`setTerm`): one-time rescan over the current rows via an injected `slotElements` provider. An unchanged term short-circuits, so a column-filter-only change never triggers a rescan.
- **Streaming** (`onAppend` / `onEvict`): `KafkaRecordsOutput.onSlotChange` updates a single slot bit per record, alongside `recordIndex`. Because those callbacks fire *before* `fireTableRowsInserted` in `ListTableModel.flushPendingAdds`, the bits are already correct when the `TableRowSorter` re-evaluates the newly inserted rows — **so free-text search stays live as records stream in, with no extra repaint hook**.
- **Slot-keyed** (not row-keyed) so the bits survive `CircularBuffer` wrap.

Per-row visibility at repaint time stays O(1): translate the row to its buffer slot via `ListTableModel.slotForRow` and consult the cached BitSet.

#### Why this shape (vs. the earlier snapshot rebuild)

An earlier iteration rebuilt the BitSet from a full `JTable` snapshot on every term change. That (a) went stale during live streaming — new rows were filtered against bits that never saw them — and (b) did O(n) string materialization on the EDT per rebuild. The incremental index fixes both: steady-state cost is one match per appended record, and the only O(n) pass is the user-initiated, debounced term-change rescan (which now walks records, not a rendered snapshot).

#### Matching surface

Free-text matches each rendered column independently (no concatenation), via the shared `cellDisplayString` helper — so typing `2026-05` hits the formatted Timestamp the user sees rather than `Date.toString()`, and a needle can't span a column boundary. Per-column substring filters compose with the free-text BitSet via `RowFilter.andFilter`. Numeric/date column operators (e.g. `offset:>100`) remain deferred — same status as the prototype; restoring them is tracked separately under plan ¶155-163.

### Stack

- PR A: CircularBuffer (#617, merged)
- PR B: ConsumerRecordIndex (#618)
- PR C: ListTableModel backed by CircularBuffer (#619)
- PR D: wire index into KafkaRecordsOutput (#620)
- **PR E: this**

### Performance

Free-text search no longer re-runs per-row `String.contains` on every repaint, nor re-materializes the whole table on every keystroke. Steady-state streaming cost is O(1) per record; the term-change rescan is O(n) once, debounced.

**Open item:** the term-change rescan runs synchronously on the EDT (matching VS Code). Validate at 50K capacity; if it ever janks, the immutable `KafkaRecord` list snapshots by reference for a cheap off-EDT rescan.

### Test plan

- [x] New `FreeTextSlotIndexTest`: incremental set/unset on append, evict clears, wrap-reuse, clear, same-term reuse, term-change rescan
- [x] New `SearchBarControllerTest` cases: free-text filter stays live as a new matching row streams in under an active term; free-text does not match across a column boundary
- [x] Removed the orphaned `buildSearchBitSet` helper + test (superseded by the index)
- [x] `./gradlew test` — full suite green


src/io/confluent/intellijplugin/consumer/data/ConsumerRecordIndex.kt
```diff
@@ -187,24 +187,4 @@ class ConsumerRecordIndex(private val capacity: Int) {
...
 }
 
-/**
- * Build a BitSet flagging buffer slots whose value matches [matcher].
- *
- * Reads [CircularBuffer.head] / `size` / `capacity` non-atomically; callers must serialize this
- * with buffer mutation. The production search path takes an EDT-side snapshot before handing off
- * to a worker thread — this helper is intended for that single-threaded code path and for tests.
- */
-fun <T : Any> buildSearchBitSet(buffer: CircularBuffer<T>, matcher: (T) -> Boolean): BitSet {
```
noeldevelops, line 197: This was a bit of YAGNI leftover from #618's original design intent, but ultimately not needed. We're building a better search bitset in the new FreeTextSlotIndex class now.

src/io/confluent/intellijplugin/consumer/editor/KafkaRecordsOutput.kt
```diff
@@ -49,38 +51,69 @@ class KafkaRecordsOutput(val project: Project, val isProducer: Boolean) : Dispos
...
         onSlotChange = ::onSlotChange,
-    ) { data, index ->
-        when (index) {
-            0 -> data.topic
-            1 -> Date(data.timestamp)
-            2 -> data.keyText ?: KafkaMessagesBundle.message("error.output.row.key")
-            3 -> data.valueText ?: data.errorText
-            4 -> data.partition
-            5 -> if (isProducer) data.duration else data.offset
-            else -> ""
```
noeldevelops, line 65: moved to `private fun columnValue` below

src/io/confluent/intellijplugin/consumer/search/SearchBarController.kt
```diff
@@ -40,6 +44,10 @@ class SearchBarController(
     private val table: JTable,
     private val filterHeader: TableFilterHeader,
     isProducer: Boolean,
+    // Star-projected: the controller only reads the live bitset and sets the term, never feeds
```
noeldevelops, line 47: in this PR I learned about [Kotlin `star projection`](https://kotlinlang.org/docs/generics.html#star-projections) ⭐ 

This controller only ever touches the read side, so it doesn't need to know the element type and `<*>` keeps it decoupled from `KafkaRecord`

PR #632, opened 2026-06-26
fix: guard ConnectionSettingsBase serialization against unloadable classes (hotfix)

noeldevelops:
## What

Hardens the **serialization** (write) path in `ConnectionSettingsBase` so a single connection property that cannot be Java-serialized is skipped (and `logger.warn`-logged) instead of aborting the entire settings save.

## Why — INTELLIJ-PLUGIN-S4

`packData` serializes each connection property via `encode64(value)`, which does a raw `ObjectOutputStream(...).writeObject(value)`. When a value's class graph references `com.intellij.ssh.config.unified.SshConfig` — not loadable on the plugin classloader in IDEs lacking the SSH module — `writeObject` throws `NoClassDefFoundError` (wrapping `ClassNotFoundException`) while introspecting the class. That escapes `getState()` and the platform reports it as:

> Exception: Cannot get ConfluentIntellijKafkaGlobalSettings component state

The read/deserialization path was already defensive (`PluginObjectInputStream` swallows `ClassNotFoundException`/`LinkageError`); the write path was not. This is the asymmetry the Sentry triage identified.

Fixes INTELLIJ-PLUGIN-S4

## Fix

Wrap the per-property `encode64` call in `packData` in a `try { } catch (t: Throwable)` **scoped to the single property** (covers `NoClassDefFoundError` / `LinkageError` / `NotSerializableException`). The catch is intentionally narrow so it cannot hide unrelated serialization bugs in other properties.

## Stacked on #631

This PR is **stacked on top of #631** (`ncothren/hotfix-t4-ssh-noclassdef`) and is the **serialization-side counterpart** to #631's deserialization guard — both changes live in the same file (`ConnectionSettingsBase.kt`), so basing this branch on #631 shares those changes and avoids conflicts. Base is set to `ncothren/hotfix-t4-ssh-noclassdef`; GitHub will auto-retarget to the release branch once #631 merges.

## Test

Added `ConnectionSettingsBasePackDataTest`: a `ConnectionData` subclass holds one property whose `writeObject` throws `NoClassDefFoundError` plus a normal `String` property. Asserts `packData` skips the exploding property and still retains the good one. Passes locally.


noeldevelops:
fixed by #631 instead

PR #594, opened 2026-04-10
[v253.25910.x] fix: Schema Registry NPE when producing Avro messages with references

bcscc:
## Summary of Changes

- Fix `NullPointerException` in `KafkaRegistryUtil.loadSchema()` when producing Avro messages whose schema has references to other schemas
  - During the `BaseClusterDataManager` refactoring (CCloud support), `loadSchema()` was changed to use unconfigured schema providers that lack a `SchemaVersionFetcher`, causing NPE in `AbstractSchemaProvider.resolveReferences()`. The fix delegates to `dataManager.parseSchemaForDisplay()` which uses properly configured providers from the `ConfluentRegistryClient`.
- Minor: clean up fully-qualified `ParsedSchema` import in `KafkaDataManager`

Fixes #591 

https://github.com/user-attachments/assets/0b426405-7ab4-4a6d-8a33-c5ab548b3072

### Click-testing instructions

1. Run a local Kafka + Schema Registry (e.g., via Docker)
2. Register an Avro schema **with references** (e.g., a `User` schema that references an `Address` schema)
3. Open the plugin, connect to the local cluster, and produce an Avro message using the referencing schema
4. Verify the message is produced successfully without errors
5. Also verify producing with a **simple** Avro schema (no references) still works

## Pull request checklist

Please check if your PR fulfills the following (if applicable):

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing

PR #636, opened 2026-07-16
fix: resolve schema references when deserializing CCloud records by schema ID

francis-a:
## Problem

Consuming records from a Confluent Cloud topic with the Schema Registry value type fails when the value schema **has schema references**. The value column shows:

```
com.google.protobuf.Descriptors$DescriptorValidationException:
  <MessageType>.<field>: ".<referenced.package.Type>" is not defined.
```

This is a follow-up to #594 / #599, which fixed reference resolution on the schema display/update paths (`CCloudClusterDataManager.parseSchemaForDisplay`) but not on the consume path.

## Root cause

`CCloudConsumerClient.fetchAndParseSchema` parses schemas fetched by ID/GUID with the two-argument `KafkaRegistryUtil.parseSchema(schemaType, schema)` overload, which defaults `references = emptyList()` and passes no resolved references. The `/schemas/ids/{id}` response includes the reference list (`SchemaByIdResponse.references`), but it was dropped.

As a result, the `ProtobufSchema` is built with no dependency files, and `toDescriptor()` in `deserializeProtobuf` cannot link imported types from referenced subjects. The same applies to Avro schemas with references.

## Fix

Mirror what `parseSchemaForDisplay` already does:

- convert `response.references` via `toSchemaReferences()`
- resolve them (including transitive references) with the existing `fetchResolvedReferences(fetcher, references)`
- parse with the four-argument `KafkaRegistryUtil.parseSchema(type, text, references, resolvedReferences)` overload introduced in #599

## Testing

- Built the plugin locally and installed it in IntelliJ 2026.1.4.
- Consumed a Confluent Cloud topic whose Protobuf value schema references another subject (which itself has transitive references): previously failed with `DescriptorValidationException`, now deserializes and renders correctly.
- Topics with reference-free schemas are unaffected


noeldevelops (approved):
Thanks for the fix! Code changes look good, reuses existing patterns ✔️  I am planning a patch release in the next week and can roll this in.

Issue #13, opened 2025-09-19
[IJPL-154088] MINOR: Kafka: Add MessagePack serializer


bcscc:
**Original Issue:** https://youtrack.jetbrains.com/issue/IJPL-154088

**Created:** 24 Jan 2023 15:03
**Priority:** Minor
**Type:** Task
**State:** Backlog
**Affected Versions:** Not specified
**Affected IDEs:** No affected ides

---

https://www.programmersought.com/article/696110287850/

---

**Migrated:** 2025-09-19

YeeaaahMan:
I would try to implement this feature.

I did similar for Kafka UI.

PR #646, opened 2026-08-30
feat: support MessagePack serializer [IJPL-154088]

YeeaaahMan:
## Summary of Changes

Added [MessagePack](https://msgpack.org/) data format support for Consumer and Producer.

- msgpack-core (org.msgpack) is the official reference implementation of the MessagePack spec maintained by the MessagePack project. Unlike moshipack, it is actively maintained and provides a full MessagePacker/MessageUnpacker API.
- jackson-dataformat-msgpack gives us a Jackson ObjectMapper(MessagePackFactory()), so payloads are serialized/deserialized through a well-known JSON-style data model. Because the same mapper is used for both serialize and validate, the two are guaranteed to agree on precedence/ECMA-404 data rules (rejecting bare tokens like nil/float), unlike the previous hybrid Gson-based validation.

1. Deploy latest Kafka-UI app using docker and [dev.yaml](https://github.com/kafbat/kafka-ui/blob/main/.dev/dev.yaml) compose file.
2. Create a new topic and add several messages using MessagePack data format (for value).
3. Check messages in the the Confluent Kafka plugin with Consumer (select Value=MessagePack + "from the beginning").
4. Similar steps for Producer.

Issue: [IJPL-154088] MINOR: Kafka: Add MessagePack serializer  https://github.com/confluentinc/intellij/issues/13 

- MessagePack serializer/deserializer rewritten around ObjectMapper(MessagePackFactory()) for consistent, spec-compliant encode/decode.
- Consumer decoding: MessagePack record values are now decoded to JSON for display in the UI.
- JSON display: format JSON to pretty for Details tab only, but compact for Data tab (table view).
- Empty payload handling: blank MessagePack input is serialized to an empty byte array.
- Block empty produce: producing a blank MessagePack payload is rejected in the producer editor.
- Stricter validation: MessagePack input is validated with Jackson (rejecting bare tokens) instead of lenient Gson.
- Friendly error messages: invalid/truncated MessagePack payloads surface a localized error (error.output.row.messagepack.invalid) instead of a raw stack trace.
- Expanded test coverage: round-trip (null/bool/number/string/array), empty/blank serialize, and empty/truncated deserialize error tests.

image

## Pull request checklist

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing


test/io/confluent/intellijplugin/common/editor/KafkaEditorUtilsTest.kt
```diff
@@ -90,4 +92,29 @@ class KafkaEditorUtilsTest {
...
+
+        @Test
+        fun `should decode messagepack byte array to compact json string`() {
+            val json = """{"a":1,"b":"two"}"""
+
+            val result = KafkaEditorUtils.getValueAsString(
+                KafkaFieldType.MESSAGEPACK, json.toByteArray(), KafkaRegistryFormat.UNKNOWN
+            )
+
+            // Decoded to compact JSON (no indentation or newlines) via tryFormatJsonCompact.
```
YeeaaahMan: remove comment 
```suggestion
```

test/io/confluent/intellijplugin/common/editor/KafkaEditorUtilsTest.kt
```diff
@@ -90,4 +92,29 @@ class KafkaEditorUtilsTest {
...
+            assertEquals(json, result)
+        }
+
+        @Test
+        fun `should return empty string for null value`() {
+            val result = KafkaEditorUtils.getValueAsString(
+                KafkaFieldType.MESSAGEPACK, null, KafkaRegistryFormat.UNKNOWN
```
YeeaaahMan, line 125: Probably, I should remove this test case.

Kafka topic can store an empty message, in theory. But during implementing [similar feature](https://github.com/kafbat/kafka-ui/pull/1741) I decided to remove ability to Produce empty messages, because it can break something suddenly when Consumer receive empty byte array.
I added this check to the Producer's field validator.

YeeaaahMan:
@noeldevelops, my PR is ready for review. 
Any feedback is welcome.

noeldevelops:
@YeeaaahMan thanks for the ping I was ooo and missed this before. I will be able to review tomorrow & I'll let you know.

src/io/confluent/intellijplugin/common/editor/KafkaEditorUtils.kt
```diff
@@ -434,11 +440,13 @@ object KafkaEditorUtils {
         }
     }
 
-    fun tryFormatJson(text: String, shemaName: String? = null): String {
+    fun tryFormatJson(text: String, shemaName: String? = null, pretty: Boolean = false): String {
```
noeldevelops, line 443: seems like this will regress for the other call sites where pretty was previously not passed?

I can't tell if that is an intentional change from your PR desc, but if so it's a behavior change to other non-MessagePack formats. If it is *intentional* for all formats, it should be called out explicitly in the PR description and given test coverage. If it's not intentional, the missing call sites should pass `pretty = true` to preserve existing behavior.

noeldevelops (changes requested):
I flagged the one thing I think is worth fixing in the `KafkaEditorUtils` file. 

A couple of minor nitpicks that I don't think are blockers but you may want to update: 

1. Empty/blank payload asymmetry between serialize and deserialize
- `MessagePackSerializer.serialize()` (line 12): blank input → returns `ByteArray(0)`.
- `MessagePackDeserializer.deserialize()`: no matching empty-array case — a 0-length payload throws `SerializationException("Invalid MessagePack payload")`.
2. Inconsistent error handling in `MessagePackSerializer.serialize()`
`serialize()` calls `jsonMapper.readTree(data)` with no try/catch, while `deserialize()` wraps failures in the new friendly, localized `SerializationException`. Malformed JSON reaching `serialize()` outside the producer UI's own pre-validation would throw a raw exception instead of a friendly message.

src/io/confluent/intellijplugin/common/editor/KafkaEditorUtils.kt
YeeaaahMan: I did it on purpose. 

If you open Producer and paste a message with several whitespaces before first quote (or bracket, or between items), it also appear and be visible on Data tab. As for me, it looks like a small bug or defect. 
(take a look at the Value column)
image

This [change](https://github.com/confluentinc/intellij/pull/646/changes/c71a33be0e6f7c6b190584765e52ad42cf11b1da) makes compact JSON representation for all types of supported messages on Data tab. And only on Details tab JSON representation still contains line breaks and indents. 
Screenshot From 2026-09-23 22-00-52

So this function is called with default value only in KafkaEditorUtils.kt and affects JSON representation only in table view.

YeeaaahMan:
>     1. Empty/blank payload asymmetry between serialize and deserialize
>     * `MessagePackSerializer.serialize()` (line 12): blank input → returns `ByteArray(0)`.
Yes, empty messages should be represented somehow in the table. 
Let it be just an empty cell. And you can export a CSV and find all of them as empty cell.

>     * `MessagePackDeserializer.deserialize()`: no matching empty-array case — a 0-length payload throws `SerializationException("Invalid MessagePack payload")`.
I don't think we should support producing of messages with empty body. Someone will use Plugin on production environment and one day will send an empty message and will brake something.
Let's add symmetry later, if someone really request it. I think, for now it should be forbidden.

>     2. Inconsistent error handling in `MessagePackSerializer.serialize()`
>        `serialize()` calls `jsonMapper.readTree(data)` with no try/catch, while `deserialize()` wraps failures in the new friendly, localized `SerializationException`. Malformed JSON reaching `serialize()` outside the producer UI's own pre-validation would throw a raw exception instead of a friendly message.
I expected, pre-validation on the UI should be enough. I will check other cases and support them.

src/io/confluent/intellijplugin/common/editor/KafkaEditorUtils.kt
noeldevelops: thanks for the clarification. Good change.

noeldevelops (approved):
thanks for the contribution and your comments! 
This is ready for merge imo, with the understanding that the SerializationException is low-risk. 

To set expectations - we don't have an official release for this plugin scheduled, but the plan is to deploy a new version before the end of October. In the meantime you're welcome to run the latest code from a local build or pre-release version of course.

noeldevelops:
I ran the pipeline locally to unblock flaky semaphore non-prod runs; confirmed `build` and `test` are passing.

